PR #66, opened 2015-06-10
Marsaud issue176

marsaud:


src/xapi.js
```diff
@@ -496,6 +498,19 @@ export default class Xapi extends XapiBase {
     )
   }
 
+  async rollingSnapshotVm (vmId, tag, depth) {
+    const vm = await this._getOrWaitObject(vmId)
+    const reg = new RegExp('^rollingSnapshot-' + tag)
```
julien-f: Special characters in `tag` must be escaped to avoid interfering with the regular expression ([escape-string-regexp](https://github.com/sindresorhus/escape-string-regexp)).

src/xapi.js
```diff
@@ -496,6 +498,19 @@ export default class Xapi extends XapiBase {
...
+    const snapshots = sortBy(filter(vm.$snapshots, snapshot => reg.test(snapshot.name_label)), 'name_label')
+    const date = new Date().toISOString()
+    const ref = await this._snapshotVm(vm, `rollingSnapshot-${tag}-${date}-${vm.name_label}`)
+    while (snapshots.length > depth - 1) {
+      const oldSnap = snapshots.shift()
+      await this.deleteVm(oldSnap.UUID)
```
julien-f: This could be parallelized by putting the promises in an array and waiting for all (`await Promise.all(promises)`).

The snapshot before the while should also be parallelized.

src/xapi.js
```diff
@@ -496,6 +498,19 @@ export default class Xapi extends XapiBase {
     )
   }
 
+  async rollingSnapshotVm (vmId, tag, depth) {
+    const vm = await this._getOrWaitObject(vmId)
```
julien-f: `_getOrWaitObject()` will wait for the object to appear if it is not present in the collection yet, this could be forever if the identifier is incorrect.

I think here it is better to fail immediately if the object does not exit and therefore `getObject()` should be used instead.

src/api/vm.coffee
```diff
@@ -654,6 +654,24 @@ exports.snapshot = snapshot
 
 #---------------------------------------------------------------------
 
+rollingSnapshot = $coroutine ({vm, tag, depth}) ->
+  snapshot = yield @getXAPI(vm).rollingSnapshotVm(vm.ref, tag, depth)
+  return snapshot.$id
+
+rollingSnapshot.params = {
+  id: { type: 'string'}
+  name: {type: 'string'}
```
julien-f: `name` or `tag`?

src/api/vm.coffee
```diff
@@ -654,6 +654,24 @@ exports.snapshot = snapshot
...
+  depth: {type: 'string'}
+}
+
+rollingSnapshot.resolve = {
+  vm: ['id', 'VM', 'administrate']
+}
+rollingSnapshot.permission = 'admin'
```
julien-f: User permissions should only be used when ACLs (via `resolve`) are not applicable.

src/api/scheduler.js
```diff
@@ -0,0 +1,16 @@
+const schedules = []
+let id = 0
+
+export async function getAll () {
+  return await Promise.resolve(schedules)
```
julien-f: I do not understand the `await Promise.resolve()` pattern.

src/xapi.js
marsaud: I feel I'll just parallelize the deletes in the while. I have to wait for the vm infos before snapshoting, and I must wait for the snapshot to succeed before starting to delete older ones (for depth 1 security)

src/api/vm.coffee
marsaud: tag ^^

src/api/scheduler.js
marsaud: Neither do I :p

src/api/job.js
```diff
@@ -0,0 +1,17 @@
+export async function getAll () {
+  return await this._jobs.get()
```
julien-f: Avoid accessing directly the underlying data, create proxy methods in `xo.js` (similar to users management).

src/api/schedule.js
```diff
@@ -0,0 +1,17 @@
+export async function getAll () {
+  return await this._schedules.get()
```
julien-f: Idem, create proxy methods in Xo.

src/xapi.js
```diff
@@ -507,6 +510,24 @@ export default class Xapi extends XapiBase {
...
+    const vm = this.getObject(vmId)
+    const reg = new RegExp('^rollingSnapshot_[^_]+_' + escapeStringRegexp(tag))
+    const snapshots = sortBy(filter(vm.$snapshots, snapshot => reg.test(snapshot.name_label)), 'name_label')
+    const date = new Date().toISOString()
+
+    const ref = await this._snapshotVm(vm, `rollingSnapshot_${date}_${tag}_${vm.name_label}`)
+
+    const promises = []
+    while (snapshots.length > depth - 1) {
+      const oldSnap = snapshots.shift()
```
julien-f: It's a detail, but accessing a `snapshot.length` and `shift()`ing at each turn of the loop is inefficient, it would probably be best to simply iterate with a `for` loop from `i = length - depth - 1` to `n = length`.

src/api/job.js
```diff
@@ -0,0 +1,17 @@
...
+}
```
julien-f: All methods should have authorization metadata.

Either `permission = 'admin'` to restrict the use to an XO administrator or a fine grained check with the `resolve` entry.

marsaud: Shouldn't we simply implement a user_id check for set() & delete() ?
I guess that the real point on authorization is when jobs will be executed on XO objects under ACLs, no ?...at this point, the methods tragetted by the job will resolve it's own permissions.

marsaud: Of course, admins may have the right to set or delete everything, but if we set .permission to "admin", what about the object owner ?...

julien-f: Administrators can do everything, even if no permissions is specified.
But if neither `permission` nor `resolve` with an ACL is defined, then everybody can call the method which poses a security problem.

Unfortunately, the ACLs do not work on objects which do not originate from Xen.
A possible solution is to check for permission yourself in the function and throwing `Unauthorized` if necessary.

marsaud: Ok, so I remove `permission` and `resolve` so anyone can 'try' the method and I check `user_id` inside the function(s)

julien-f: Yep.

src/api/vm.coffee
```diff
@@ -486,6 +486,23 @@ exports.snapshot = snapshot
 
 #---------------------------------------------------------------------
 
+rollingSnapshot = $coroutine ({vm, tag, depth}) ->
+  snapshot = yield @getXAPI(vm).rollingSnapshotVm(vm.ref, tag, depth)
+  return snapshot.$id
+
```
julien-f: Please, add a small `rollingSnapshot.description` which will be expose by introspection to xo-cli.

src/api/job.js
```diff
@@ -0,0 +1,33 @@
+// FIXME so far, no acls for jobs
+
+export async function getAll () {
+  return await this.getAllJobs()
+}
+
```
julien-f: Please, add a `<fn>.description` to all API methods.

These are exposed via introspection to xo-cli.

src/models/job.js
```diff
@@ -0,0 +1,48 @@
...
+
+  get idPrefix () {
+    return 'job:'
+  }
+
+  async create (user_id, job) {
+    job.user_id = user_id
+    // Serializes.
+    job.paramsVector = JSON.stringify(job.paramsVector)
+    return this.add(new Job(job))
```
julien-f: Isn't it more explicit to say `return await this.add...`?

src/models/schedule.js
```diff
@@ -0,0 +1,33 @@
...
+export class Schedules extends Collection {
+  get Model () {
+    return Schedule
+  }
+
+  get idPrefix () {
+    return 'schedule:'
+  }
+
+  create (user_id, job, cron, autoEnable) {
```
julien-f: We should use camelCasing in new code (`userId`).

src/api/job.js
```diff
@@ -0,0 +1,103 @@
...
+						}
+					}
+				}
+			}
+		}
+	}
+}
+
+export async function set ({job}) {
+  return await this.updateJob(job)
```
julien-f: The `set()` method does not return anything (`true` in fact due to the implementation requirement), can you remove the `return` to avoid polluting the return value?

src/api/job.js
```diff
@@ -0,0 +1,103 @@
...
+						}
+					}
+				}
+			}
+		}
+	}
+}
+
+async function delete_ ({id}) {
+  return await this.removeJob(id)
```
julien-f: Idem: no `return`.

src/api/schedule.js
```diff
@@ -0,0 +1,55 @@
...
+set.description = 'Modifies an existing schedule'
+set.params = {
+	id: {type: 'string'},
+	jobId: {type: 'string', optional: true},
+	cron: {type: 'string', optional: true},
+	enabled: {type: 'boolean', optional: true}
+}
+
+async function delete_ ({id}) {
+  return await this.removeSchedule(id)
```
julien-f: No `return`.

src/index.js
```diff
@@ -435,7 +448,9 @@ export default async function main (args) {
   connect.use(bind(xo._handleProxyRequest, xo))
 
   // Must be set up before the static files.
-  setUpApi(webServer, xo)
+  const api = setUpApi(webServer, xo)
+
+  const scheduler = setUpScheduler(api, xo)
```
julien-f: Where is this variable used?

It is weird not to have a lint error.

julien-f: My bad, I saw it.

src/job-executor.js
```diff
@@ -0,0 +1,74 @@
...
+  assign(product, ...args)
+  return product
+}
+
+export default class JobExecutor {
+  constructor (xo, api) {
+    this.xo = xo
+    this.api = api
+    this._extractValueCb = {
+      'set': items => items.values
```
julien-f: Quotes are superfluous here.

src/job-executor.js
```diff
@@ -0,0 +1,74 @@
...
+      paramsFlatVector = this._computeCrossProduct(paramsVector.items, productParams, this._extractValueCb)
+    } else {
+      throw new UnsupportedVectorType(paramsVector)
+    }
+    const connection = this.xo.createUserConnection()
+    connection.set('user_id', userId)
+    forEach(paramsFlatVector, params => {
+      const params_ = {}
+      assign(params_, params)
+      this.api.call(connection, method, params_)
```
julien-f: You could inline the assign: `this.api.call(connection, method, assign({}, params)`

src/job-executor.js
marsaud: Without quotes, the linter rises a "missing getter" warning...

PR #70, opened 2015-06-18
vm.create() refactoring

julien-f:


julien-f:
There are still some tests to do but at least it is less bugged than the current version and provides some other fixes.

PR #75, opened 2015-07-02
Scheduled VM Backup feature

marsaud:


src/api/vm.coffee
```diff
@@ -636,6 +638,29 @@ exports.export = export_;
 
 #---------------------------------------------------------------------
 
+exportToPath = $coroutine ({vm, path, compress}) ->
+  targetStream = fs.createWriteStream(path)
+  sourceStream = yield @getXAPI(vm).exportVm(vm.id, compress ? true)
+  sourceStream.pipe(targetStream)
+  yield eventToPromise(targetStream, 'finish')
+  return
```
julien-f: Security issue: if the user as the administration on any VM or snapshot it can write anywhere in the file system with the same rights as xo-server (`root` on XOA) which means it can potentially destroy the system!

I would much prefer to have a base path set in the configuration file or even better a base path and auto generated file names.

marsaud: Ok for both, file names will be auto generated in the same way than rolling snapshot names.

For the basePath config, you think about the yaml config file ?

julien-f: Yes, maybe under the name `backupsPath`?

marsaud: See my suggestion in next commit. It's as you want

julien-f: I think it would be better to group entries related to backups than grouping paths (i.e. `backup.path`).

src/mounter.js
```diff
@@ -0,0 +1,102 @@
...
+import forEach from 'lodash.foreach'
+import fs from 'fs-extra'
+import {exec} from 'child_process'
+
+const execAsync = Bluebird.promisify(exec)
+Bluebird.promisifyAll(fs)
+
+export default class Mounter {
+  constructor (xo) {
+    this.xo = xo
```
julien-f: `Mounter` does not need to keep a reference to `xo`.

src/api/mount.js
```diff
@@ -0,0 +1,60 @@
...
+  await this.removeMount(id)
+}
+
+delete_.permission = 'admin'
+delete_.description = 'Deletes an existing nfs mount point'
+delete_.params = {
+	id: {type: 'string'}
+}
+
+export function getBackupMountPath () {
```
julien-f: If we remove the root path, this method is no longer necessary.

src/api/mount.js
```diff
@@ -0,0 +1,60 @@
...
+  return await this.getMount(id)
+}
+
+get.permission = 'admin'
+get.description = 'Gets an existing nfs mount point'
+get.params = {
+	id: {type: 'string'}
+}
+
+export async function create ({path, host, share}) {
```
julien-f: Replace `path` by a human `name`.

Replace `host` and `share` by a proper URL:
- `nfs://`
- `file://`

sample.config.yaml
```diff
@@ -102,3 +102,13 @@ redis:
...
+
+#=====================================================================
+
+# Backup configuration
+backup:
+  # Base path for scheduled VM backups
+  # Default value: /mnt/xoa/backup
+  # Warning : all mount points and backups managed through xoa are relative to this path
+  #   so on change, mount target paths and backup paths will be moved accordingly
+  #'path': '/mnt/xoa/backup'
```
julien-f: The root path is no longer necessary: the mounts should be in a volatile directory (for instance, `/tmp/xo-server/mounts`)

julien-f: A default mount `local` (`file:///var/lib/xoa-backups`) could be created at startup if there are no mounts defined.

src/xapi.js
```diff
@@ -666,6 +669,33 @@ export default class Xapi extends XapiBase {
     return await this._getOrWaitObject(ref)
   }
 
+  async rollingBackupVm (vmId, path, tag, depth, compress) {
```
julien-f: Because it access the file system, it should, IMHO, be implemented in `Xo`.

What do you think?

src/xapi.js
```diff
@@ -666,6 +669,33 @@ export default class Xapi extends XapiBase {
...
+    await fs.ensureDirAsync(path)
+    const files = await fs.readdirAsync(path)
+
+    const reg = new RegExp('^[^_]+_' + escapeStringRegexp(tag))
+    const backups = sortBy(filter(files, fileName => reg.test(fileName)))
+
+    const date = new Date().toISOString()
+    const backupFullPath = `${path}/${date}_${tag}_${vm.name_label}.xva`
+
+    const targetStream = fs.createWriteStream(backupFullPath)
```
julien-f: Use `flags` to ensure not overwriting an existing file.

src/xapi.js
julien-f: `rollingSnapshot()` is based on top `snapshot()` but `rollingBackup()` is not based on a `backup()` method, is there a reason?

src/api/vm.coffee
```diff
@@ -487,6 +490,30 @@ exports.rollingSnapshot = rollingSnapshot
 
 #---------------------------------------------------------------------
 
+rollingBackup = $coroutine ({vm, path, tag, depth, compress}) ->
```
julien-f: There should be a mount/remote id parameter.

```js
const remote = this.getRemote(remoteId)
return await this.getXAPI(vm).rollingBackupVm(vm.id, remote, path, tag, depth, compress)
```
```

src/xapi.js
marsaud: As I check the fopen man page, there is no flag to really avoid overwriting an existing file. There are flags to append instead of truncate to zero, but does this make sense for a VM export ?...

julien-f: Also in http://devdocs.io/iojs/fs#fs_fs_open_path_flags_mode_callback

src/xapi.js
marsaud: backup() and rollingBackup() will exist in xo, as the simple backup itself introduces the access to the file system.

PR #76, opened 2015-07-09
Deleting a VM also deletes its snapshots

marsaud:


src/xapi.js
```diff
@@ -592,6 +592,12 @@ export default class Xapi extends XapiBase {
       }))
     }
 
+    await Promise.all(map(vm.$snapshots, snapshot => {
+      try {
+        return this.deleteVm(snapshot.uuid, true).catch(err => {console.log('err', err);noop})
+      } catch (_) {console.log('_', _.message)}
+    }))
+
```
julien-f: I wonder if it should be optional via a boolean like `deleteSnapshots`?

marsaud: I wondered too, but according to @olivierlambert , a orphan snapshot is useless...

src/xapi.js
```diff
@@ -592,6 +592,12 @@ export default class Xapi extends XapiBase {
       }))
     }
 
+    await Promise.all(map(vm.$snapshots, snapshot => {
+      try {
```
julien-f: `deleteVm()` is asynchronous, it cannot throw synchronously an exception.

src/xapi.js
```diff
@@ -592,6 +592,12 @@ export default class Xapi extends XapiBase {
...
+        return this.deleteVm(snapshot.uuid, true).catch(err => {console.log('err', err);noop})
```
julien-f: Use `$id` instead of `uuid`.

julien-f: `noop` is useless here.

PR #83, opened 2015-09-11
Tags

olivierlambert:
This PR is not perfect and should be reviewed by @julien-f 

I had problems with `getXAPI` which should allow me to have the `ref` of the object directly, but it didn't work (IDK why).

Thus, that's not very nice to write:

`this.getXAPI(id).call(this.getObject(id).type+'.add_tags', this.getObject(id).ref, tag)`


src/api/tags.js
```diff
@@ -0,0 +1,31 @@
+import {InvalidParameters} from '../api-errors'
+
+// ===================================================================
+
+export async function addTag ({tag, id}) {
+
+  return await this.getXAPI(id).call(this.getObject(id).type+'.add_tags', this.getObject(id).ref, tag)
```
olivierlambert: This should be improved

src/api/tags.js
```diff
@@ -0,0 +1,31 @@
...
+addTag.permission = 'admin'
+
+addTag.params = {
+  tag: { type: 'string' },
+  id: { type: 'string' }
+}
+
+export async function removeTag ({tag, id}) {
+
+  return await this.getXAPI(id).call(this.getObject(id).type+'.remove_tags', this.getObject(id).ref, tag)
```
olivierlambert: Same stuff here

src/api/tags.js
```diff
@@ -0,0 +1,31 @@
+import {InvalidParameters} from '../api-errors'
+
+// ===================================================================
+
+export async function add ({tag, id}) {
+
+  return await this.getXAPI(id).call(this.getObject(id).type+'.add_tags', this.getObject(id).ref, tag)
```
olivierlambert: Ugly :/

src/api/tags.js
```diff
@@ -0,0 +1,31 @@
...
+add.permission = 'admin'
+
+add.params = {
+  tag: { type: 'string' },
+  id: { type: 'string' }
+}
+
+export async function remove ({tag, id}) {
+
+  return await this.getXAPI(id).call(this.getObject(id).type+'.remove_tags', this.getObject(id).ref, tag)
```
olivierlambert: Same stuff

src/api/tags.js
julien-f: Why `return`? What value are you returning?

You should implement the logic in `xapi.js` and simply use it from the API.

src/api/tags.js
```diff
@@ -0,0 +1,31 @@
...
+// ===================================================================
+
+export async function add ({tag, id}) {
+
+  return await this.getXAPI(id).call(this.getObject(id).type+'.add_tags', this.getObject(id).ref, tag)
+}
+
+add.description = 'add a new tag into an object'
+
+add.permission = 'admin'
```
julien-f: Why not using ACLs?

src/api/tags.js
olivierlambert: Void. If you want to make some changes, go ahead, I won't be able to do them (I will waste too many time to make it work otherwise). The goal of this PR review is to shape it like you wanted and merge it after :dancer:

PR #85, opened 2015-09-14
Password change API for any user. Related to xo-web issue#362

marsaud:


src/xo.js
```diff
@@ -334,6 +334,22 @@ export default class Xo extends EventEmitter {
     })
   }
 
+  async changePassword (oldPassword, newPassword) {
+    const id = this.session.get('user_id')
```
julien-f: `this.session` is not defined on Xo, it only make sense in the context of an API method, therefore this value should be fetched there and passed a parameter.

src/xo.js
```diff
@@ -334,6 +334,22 @@ export default class Xo extends EventEmitter {
...
+    const user = await this._getUser(id)
```
julien-f: `Xo#_getUser()` returns an instance of User and not a plain object, therefore its properties should be fetched using the `User#get()` method.

PR #105, opened 2015-10-26
New parser for host and vms stats.

Wescoeur:
The data are cached for other http requests.


src/api/host.coffee
```diff
@@ -283,97 +283,20 @@ installAllPatches.resolve = {
...
-    load: load
-    timestamp: Date.now()
-  }
+  # granularity:
+  # - 0: every 5 sec along last 10 minutes
+  # - 1: every minute along last 2 hours
+  # - 2: every hour along past week
+  # - 3: everyday along past year
+  # see http://xenserver.org/partners/developing-products-for-xenserver/18-sdk-development/96-xs-dev-rrds.html
+  granularity = if granularity then granularity else 0
```
julien-f: Maybe too low level/opaque, would not a string (e.g. `seconds`, `minutes`, `hours`, `days`) be better?

src/xo.js
```diff
@@ -908,6 +912,14 @@ export default class Xo extends EventEmitter {
     return xapi
   }
 
+  getXapiVmStats (hostname, granularity, sessionId, vmId) {
```
julien-f: Hide low level stuff, just pass `vmId` and `granularity` and let `Xo#getXapiVmStats()` grabs the others.

src/xo.js
```diff
@@ -908,6 +912,14 @@ export default class Xo extends EventEmitter {
...
+    return this._xapiStats.getVmPoints(hostname, granularity, sessionId, vmId)
+  }
+
+  getXapiHostStats (hostname, granularity, sessionId) {
```
julien-f: Idem.

PR #106, opened 2015-10-27
'Pending tasks' panel: upload task displays patch name

pdonias:


julien-f:
`Xapi#uploadPoolPatch()` is also used in [pool.js#L59](https://github.com/vatesfr/xo-server/blob/next-release/src/api/pool.js#L59).

PR #116, opened 2015-11-06
vm.copy() can a copy a VM on a remote or local SR.

julien-f:


src/xapi.js
```diff
@@ -568,23 +578,40 @@ export default class Xapi extends XapiBase {
...
+      await this._copyVm(
+        this.getObject(vmId),
+        nameLabel,
+        this.getObject(srId)
+      )
     )
   }
 
+  async remoteCopyVm (vmId, srXapi, srId) {
+    const sr = srXapi.getObject(sr)
```
marsaud: `const sr = srXapi.getObject(srId) ?`

julien-f: Indeed, thanks :)

PR #118, opened 2015-11-06
Disaster recovery feature

marsaud:


src/xo.js
```diff
@@ -796,6 +796,41 @@ export default class Xo extends EventEmitter {
...
+        if (vm && reg.test(vm.name_label)) {
+          vms.push(vm)
+        }
+      }
+    })
+    const olderCopies = sortBy(vms)
+
+    const sourceXapi = this.getXAPI(vm)
+    const targetXapi = this.getXAPI(sr)
+    const drCopy = await sourceXapi.remoteCopyVm(vm, targetXapi, sr)
```
julien-f: If I am not mistaken, it should be `vm.id` and `sr.id`.

src/xo.js
```diff
@@ -796,6 +796,41 @@ export default class Xo extends EventEmitter {
...
+
+    const date = safeDateFormat(new Date())
+    const copyName = `${date}_${tag}_${vm.name_label}`
+
+    await targetXapi.call('VM.set_name_label', drCopy.ref, copyName)
```
julien-f: Avoid calling directly the XAPI, implement a basic `Xapi#setVmProperties()` instead.

julien-f: Otherwise, I can implement this functionality directly in `Xapi#remoteCopyVm()`.

marsaud: Why not, but I saw a heavy use of this in vm.coffee ... I would prefer take time specifically on a clean setVmProperties with a global API code review, than doing something basic on the fly for the DR feature...but I'll follow your final word.

src/xo.js
```diff
@@ -796,6 +796,41 @@ export default class Xo extends EventEmitter {
...
+    await targetXapi.addTag(drCopy.id, 'Disaster Recovery')
```
julien-f: I now it's far too messy now, but `drCopy` does is a `xen-api` object (not a XO object`), consequently it does not have an `id` property, but it does have `$id`.

src/xo.js
julien-f: It's heavily used because it's legacy code, it should be updated when possible.

marsaud: Yeah, I know, I just suggest we keep this for a specific update effort ... unless you fear the effort would come late ...

marsaud:
To reopen targetting next-release

PR #119, opened 2015-11-09
Disaster recovery feature

marsaud:


src/api/vm.coffee
```diff
@@ -543,6 +543,29 @@ exports.rollingBackup = rollingBackup
 
 #---------------------------------------------------------------------
 
+rollingDRCopy = $coroutine ({vm, pool, tag, depth}) ->
+  if vm.$poolId is pool.id
+    throw new Error('Disaster Recovery attempts to copy on the same pool')
```
julien-f: You can throw `JsonRpcError()` if you want this message to be transmitted to the client.

src/api/vm.coffee
```diff
@@ -543,6 +543,29 @@ exports.rollingBackup = rollingBackup
...
+  return yield @rollingDRCopyVm({vm, sr: @getObject(pool.default_SR, 'SR'), tag, depth})
```
julien-f: `yield` just before `return` is unnecessary (therefore `$coroutine` is too for this method) and complexify the generated code, I suggest you remove them.

src/api/vm.coffee
```diff
@@ -543,6 +543,29 @@ exports.rollingBackup = rollingBackup
...
+rollingDRCopy.params = {
+  id: { type: 'string' }
+  pool: { type: 'string' }
+  tag: { type: 'string'}
+  depth: { type: 'number' }
+}
+
+rollingDRCopy.resolve = {
+  vm: ['id', ['VM', 'VM-snapshot'], 'administrate'],
+  pool: ['pool', ['pool'], 'administrate']
```
julien-f: The array is unnecessary here for the type as it is single.

src/xo.js
```diff
@@ -794,6 +794,42 @@ export default class Xo extends EventEmitter {
...
+    const drCopy = await this.getXAPI(vm).remoteCopyVm(vm.id, targetXapi, sr.id, copyName)
+    await targetXapi.addTag(drCopy.$id, 'Disaster Recovery')
+
+    const promises = []
+    for (let surplus = olderCopies.length - (depth - 1); surplus > 0; surplus--) {
+      const oldDRVm = olderCopies.shift()
+      promises.push(targetXapi.deleteVm(oldDRVm.id, true))
+    }
+
+    await Bluebird.all(promises)
```
julien-f: Use `Promise.all()` instead as it is more standard.

src/xo.js
```diff
@@ -794,6 +794,42 @@ export default class Xo extends EventEmitter {
...
+            vms.push(vm)
+          }
+        }
+      } catch (error) {
+        if (!(error instanceof NoSuchObject)) {
+          throw error
+        }
+      }
+    })
+    const olderCopies = sortBy(vms)
```
julien-f: The second argument is missing isn't it?

src/api/vm.coffee
```diff
@@ -543,6 +543,29 @@ exports.rollingBackup = rollingBackup
 
 #---------------------------------------------------------------------
 
+rollingDRCopy = ({vm, pool, tag, depth}) ->
```
julien-f: I think I would prefer `rollingDrCopy` instead as I find it easier to split the words. What is your opinion?

marsaud: DR has been a noticable acronym @olivierlambert used a lot to mention Disaster recovery feature. To be true I don't care, you choose :)

julien-f: It is an acronym but they should be camel cased as well to avoid ambiguity and inconsistency (yes, `XMLHttpRequest()`, I am looking at you -_- ).

marsaud: Yeah, it would be sad to be inconsistent with VM or SR :p

Ok, I turn all the DR methods to Dr

src/xo.js
```diff
@@ -794,6 +794,42 @@ export default class Xo extends EventEmitter {
     return backupFullPath
   }
 
+  async rollingDrCopyVm ({vm, sr, tag, depth}) {
+    tag = 'DR_' + tag
+    const reg = new RegExp('^[^_]+_' + escapeStringRegexp(tag))
+
+    const vms = []
+    forEach(this.getObjects(sr.VDIs), vdi => {
```
julien-f: You should use XAPI objects instead (`xapi.getObject()`) which are easier to use (e.g. `sr.$VDIs`) and are should not change in the future.

marsaud: Why is it easier to use ?... sr.$VDIs will already be an array of VDIs right ?

julien-f: Yes, XAPI objects are automatically linked together, therefore you don't need to use `getObject()` so often, for instance: `pool.$master.$resident_VMs[0].name_label`

julien-f:
Squash and rebase on `next-release` please.

src/xo.js
```diff
@@ -812,6 +812,36 @@ export default class Xo extends EventEmitter {
...
+    const reg = new RegExp('^[^_]+_' + escapeStringRegexp(`${tag}_${vm.name_label}`) + '$')
+
+    const targetXapi = this.getXAPI(sr)
+    sr = targetXapi.getObject(sr.id)
+    const sourceXapi = this.getXAPI(vm)
+    vm = sourceXapi.getObject(vm.id)
+
+    const vms = []
+    forEach(sr.$VDIs, vdi => {
+      const vm = vdi.$VBDs && vdi.$VBDs[0] && vdi.$VBDs[0].$VM
```
julien-f: Avoid accessing too many times $VMDs as it calls each time a function which create a new array.

Use an intermediary variable if necessary.

PR #125, opened 2015-11-13
Add VM snapshot quiesce support

olivierlambert:


src/xapi.js
```diff
@@ -586,8 +586,16 @@ export default class Xapi extends XapiBase {
   }
 
   async _snapshotVm (vm, nameLabel = vm.name_label) {
-    const ref = await this.call('VM.snapshot', vm.$ref, nameLabel)
-
+    let ref
+    try {
+      ref = await this.call('VM.snapshot_with_quiesce', vm.$ref, nameLabel)
+      throw new Error('No quiesce support, fallback to basic snapshot')
```
julien-f: I believe this error should not be there.

olivierlambert: What do you mean? Putting this piece of code elsewhere?

PR #129, opened 2015-11-18
Avoid ghost user and ghost group.

Wescoeur:


src/xo.js
```diff
@@ -45,6 +45,7 @@ import {
 } from './api-errors'
 import {Jobs} from './models/job'
 import {ModelAlreadyExists} from './collection'
+import {noop} from './utils'
```
julien-f: There is already an import from `utils`, can you group them?

PR #130, opened 2015-11-18
Enhancements to prepare generic job scheduling

marsaud:


src/index.js
```diff
@@ -442,9 +442,18 @@ const setUpApi = (webSocketServer, xo, verboseLogsOnErrors) => {
   })
 }
 
+const setUpJobExecutor = xo => {
+  const executor = new JobExecutor(xo)
+  xo.executor = executor
```
julien-f: Use `xo.defineProperty('executor', executor)` instead.

And I believe the name jobExecutor would be clearer.

src/index.js
```diff
@@ -442,9 +442,18 @@ const setUpApi = (webSocketServer, xo, verboseLogsOnErrors) => {
...
+  return executor
+}
+
 const setUpScheduler = xo => {
-  const jobExecutor = new JobExecutor(xo)
-  const scheduler = new Scheduler(xo, {executor: jobExecutor})
+  if (!xo.executor) {
+    setUpJobExecutor(xo)
+  }
+  const scheduler = new Scheduler(xo, {executor: xo.executor})
```
julien-f: Maybe it is no longer necessary to pass the job executor as it is available from `xo`.

src/xo.js
```diff
@@ -647,6 +647,28 @@ export default class Xo extends EventEmitter {
     return await this._jobs.remove(id)
   }
 
+  async runJobSequence (idSequence) {
+    const notFound = []
+    forEach(idSequence, async id => {
```
julien-f: Jobs are still run in parallel: `forEach()` does not wait for async functions to terminate before iterating the next items.

marsaud: A while (id = idSequence.pop()) like ?...

julien-f: Yep, or a `for (; ;)` or a `for ( of )`, just use a native syntax.

src/xo.js
```diff
@@ -647,6 +647,28 @@ export default class Xo extends EventEmitter {
     return await this._jobs.remove(id)
   }
 
+  async runJobSequence (idSequence) {
+    const notFound = []
+    for (let id of idSequence) {
```
julien-f: You can use `const` with `for ( of )`.

src/index.js
marsaud: Can you explain me quickly the benefit to use defineProperty in this case ? (no emergency)

julien-f: There is no major benefits.

I am just trying to provide a standard interface to extend an `xo` instance from *the outside*.

Minor benefits:
- the property cannot be overwritten
- if the property is a function, it does not have `xo` for context and cannot pollute accidentally

src/index.js
marsaud: I don't think the Scheduler class can take that for sure... Scheduler uses xo and and the executor as 2 different components. The executor is only available in xo by today's architecture choices...

julien-f: As you wish.

PR #131, opened 2015-11-19
remove tokens on user deletion

Wescoeur:


src/xo.js
```diff
@@ -341,10 +322,15 @@ export default class Xo extends EventEmitter {
 
   async deleteUser (id) {
     const user = await this.getUser(id)
+    const tokens = await this._getAuthenticationTokensForUser(id)
```
julien-f: I think it could be done asynchronously and any errors should be ignored.

src/xo.js
```diff
@@ -341,10 +322,15 @@ export default class Xo extends EventEmitter {
...
 
     await this._users.remove(id)
 
     // Remove the user from all its groups.
+    forEach(tokens, token => {
+      this._tokens.remove(token.id)
```
julien-f: This is bad!!

The promises returned by `_tokens.remove()` are ignored, therefore errors are unhandled and can crash the program.

PR #133, opened 2015-11-19
Avoid plugin loading if config is not valid

Wescoeur:


src/xo.js
```diff
@@ -1427,9 +1428,7 @@ export default class Xo extends EventEmitter {
     )
   }
 
-  async configurePlugin (id, configuration) {
-    const plugin = await this._getRawPlugin(id)
-
+  async _configurePlugin (plugin, configuration) {
```
julien-f: On success, mark the plugin as configured.

julien-f:
Mark non configurable plugin as already configured.

Check before loading that a plugin is configured.

julien-f:
Also, don't wait and ignore failures for configuration and loading in registration.

src/xo.js
```diff
@@ -1384,13 +1387,18 @@ export default class Xo extends EventEmitter {
...
 
-    if (configuration) {
-      await instance.configure(configuration)
-    }
-
-    if (autoload) {
-      await this.loadPlugin(id)
-    }
+    Promise.resolve()
+      .then(value => {
```
julien-f: This function does not expect any parameters.

src/xo.js
```diff
@@ -1384,13 +1387,18 @@ export default class Xo extends EventEmitter {
...
+        if (!plugin.configured) {
+          return this._configurePlugin(plugin, configuration)
+        }
+      })
+      .then(value => {
```
julien-f: Idem.

src/xo.js
```diff
@@ -1436,6 +1442,13 @@ export default class Xo extends EventEmitter {
 
     // Sets the plugin configuration.
     await plugin.instance.configure(configuration)
+    plugin.configured = true
+  }
+
+  async configurePlugin (id, configuration) {
+    const plugin = await this._getRawPlugin(id)
```
julien-f: No need to `await`, `_getRawPlugin()` is sync.

src/xo.js
```diff
@@ -1384,13 +1387,18 @@ export default class Xo extends EventEmitter {
...
     }
 
-    if (configuration) {
-      await instance.configure(configuration)
-    }
-
-    if (autoload) {
-      await this.loadPlugin(id)
-    }
+    Promise.resolve()
```
julien-f: I'd like a comment explaining what is done, why and how.

src/xo.js
```diff
@@ -1360,6 +1360,7 @@ export default class Xo extends EventEmitter {
     const id = name
 
     this._plugins[id] = {
```
julien-f: Move `const plugin =` here.

PR #135, opened 2015-11-19
emergencyHostShutdown(hostId) : suspends all the VMs running on the h…

pdonias:
…ost and then shuts the host down

### DIFF
--- a/src/xapi.js
+++ b/src/xapi.js
@@ -23,7 +23,8 @@ import {
   mapToArray,
   noop,
   parseXml,
-  pFinally
+  pFinally,
+  pSettle
 } from './utils'
 import {JsonRpcError} from './api-errors'
 
@@ -558,6 +559,21 @@ export default class Xapi extends XapiBase {
         host = this.getObject(host.$id)
       }
     }
+  }
+
+  async emergencyShutdownHost (hostId) {
+    const host = this.getObject(hostId)
+    const vms = host.$resident_VMs
+    debug(`Emergency shutdown: ${host.name_label}`)
+    await pSettle(
+      mapToArray(vms, vm => {
+        if (!vm.is_control_domain) {
+          return this.call('VM.suspend', vm.$ref)
+        }
+      })
+    )
+    await this.call('host.disable', host.$ref)
+    await this.call('host.shutdown', host.$ref)
   }
 
   // =================================================================
--- a/src/api/host.coffee
+++ b/src/api/host.coffee
@@ -256,6 +256,23 @@ exports.installAllPatches = installAllPatches
 
 #---------------------------------------------------------------------
 
+emergencyShutdownHost = ({host}) ->
+  return @getXAPI(host).emergencyShutdownHost(host._xapiId)
+
+emergencyShutdownHost.description = 'suspend all VMs and shutdown host'
+
+emergencyShutdownHost.params = {
+  host: { type: 'string' }
+}
+
+emergencyShutdownHost.resolve = {
+  host: ['host', 'host', 'administrate']
+}
+
+exports.emergencyShutdownHost = emergencyShutdownHost
+
+#---------------------------------------------------------------------
+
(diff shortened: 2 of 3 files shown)


src/xapi.js
```diff
@@ -560,6 +560,18 @@ export default class Xapi extends XapiBase {
     }
   }
 
+  async emergencyHostShutdown (hostId) {
```
julien-f: `emergencyShutdownHost()`.

src/xapi.js
```diff
@@ -560,6 +560,18 @@ export default class Xapi extends XapiBase {
...
+    const host = this.getObject(hostId)
+    const vms = host.$resident_VMs
+    debug('Emergency shutdown : ' + host.name_label)
+    for (let vm of vms) {
```
julien-f: You can use `const` instead of `let` here.

src/xapi.js
```diff
@@ -560,6 +560,18 @@ export default class Xapi extends XapiBase {
     }
   }
 
+  async emergencyHostShutdown (hostId) {
+    const host = this.getObject(hostId)
+    const vms = host.$resident_VMs
+    debug('Emergency shutdown : ' + host.name_label)
```
julien-f: In English, no space before colon.

julien-f: Use ES2015 template string instead.

src/xapi.js
```diff
@@ -560,6 +560,18 @@ export default class Xapi extends XapiBase {
...
+    for (let vm of vms) {
+      if (includes(vm.allowed_operations, 'suspend')) {
+        await this.call('VM.suspend', vm.$ref)
```
julien-f: No need to test if the operation is allowed, simply do it, and handle errors.

src/xapi.js
```diff
@@ -560,6 +560,18 @@ export default class Xapi extends XapiBase {
     }
   }
 
+  async emergencyHostShutdown (hostId) {
+    const host = this.getObject(hostId)
+    const vms = host.$resident_VMs
+    debug('Emergency shutdown : ' + host.name_label)
+    for (let vm of vms) {
+      if (includes(vm.allowed_operations, 'suspend')) {
```
julien-f: You should avoid the control domain VM (`is_control_domain` is truthy).

src/xapi.js
```diff
@@ -560,6 +560,18 @@ export default class Xapi extends XapiBase {
...
+        await this.call('VM.suspend', vm.$ref)
+      }
+    }
```
julien-f: Please, suspend VMs in parallel.

src/xapi.js
```diff
@@ -560,6 +560,18 @@ export default class Xapi extends XapiBase {
...
+    this.shutdownHost(hostId)
```
julien-f: After consideration, do not use `this.shutdownHost()` which evacuates VMs, use directly `this.call('host.shutdown', host.$ref)`.

src/xapi.js
```diff
@@ -560,6 +561,21 @@ export default class Xapi extends XapiBase {
     }
   }
 
+  async emergencyShutdownHost (hostId) {
+    const host = this.getObject(hostId)
+    const vms = host.$resident_VMs
+    debug(`Emergency shutdown: ${host.name_label}`)
+    const promises = []
```
julien-f: Use a map.

src/xapi.js
```diff
@@ -560,6 +561,21 @@ export default class Xapi extends XapiBase {
...
+    debug(`Emergency shutdown: ${host.name_label}`)
+    await pSettle(
+      mapToArray(vms, vm => {
+        if (!vm.is_control_domain) {
+          return this.call('VM.suspend', vm.$ref)
+        }
+      })
+    )
+    await this.call('host.disable', host.$ref)
+    this.call('host.shutdown', host.$ref)
```
julien-f: You should also wait for the shutdown to complete.

PR #136, opened 2015-11-19
Cloud config management for CoreOS

olivierlambert:


src/xapi.js
```diff
@@ -1345,6 +1345,28 @@ export default class Xapi extends XapiBase {
...
+    const vm = this.getObject(vmId)
+    const host = this.pool.$master
+
+    await this.call('host.call_plugin', host.$ref, 'xscontainer', 'create_config_drive', {
+      vmuuid: vm.uuid,
+      sruuid: srId,
+      configuration: config
+    })
+    // after creating the cloud drive, activate the Docker monitoring
+    this.registerDockerContainer(vm.uuid) // FIXME: doesn't not work? async issue?
```
olivierlambert: I have to check if I can register the Docker plugin before the VM is booted for the first time.

src/xapi.js
```diff
@@ -1345,6 +1345,28 @@ export default class Xapi extends XapiBase {
...
+    })
+  }
+
+  async createCloudConfigDrive (vmId, srId, config) {
+    const vm = this.getObject(vmId)
+    const host = this.pool.$master
+
+    await this.call('host.call_plugin', host.$ref, 'xscontainer', 'create_config_drive', {
+      vmuuid: vm.uuid,
+      sruuid: srId,
```
julien-f: You can't do that, you have to use a valid UUID.

PR #137, opened 2015-11-20
Add: Job and crossProduct schemas.

Wescoeur:


src/schemas/job.js
```diff
@@ -0,0 +1,38 @@
+export default {
+  $schema: 'http://json-schema.org/draft-04/schema#',
+  type: 'object',
+  properties: {
+    type: {
+      enum: ['call']
+    },
+    id: {
+      type: 'string',
+      description: 'job id'
```
julien-f: Not meaningful.

src/schemas/job.js
```diff
@@ -0,0 +1,38 @@
...
+    },
+    userId: {
+      type: 'string',
+      description: 'user id'
```
julien-f: You should explain what is the role of this property.

src/schemas/job.js
```diff
@@ -0,0 +1,38 @@
...
+    },
+    key: {
+      type: 'string',
+      description: ''
```
julien-f: If you don't know, leave free and write a TODO.

src/schemas/job.js
```diff
@@ -0,0 +1,38 @@
...
+    },
+    method: {
+      type: 'string',
+      description: 'called method'
+    },
+    paramsVector: {
+      oneOf: [
+        { '$ref': 'job/crossProduct.js' }
```
julien-f: Do not use a ref like this, it won't work, use a proper import.

src/schemas/job/crossProduct.js
```diff
@@ -0,0 +1,32 @@
+export default {
```
julien-f: Name `params-vector.js` instead.

src/schemas/job/crossProduct.js
```diff
@@ -0,0 +1,32 @@
...
+  $schema: 'http://json-schema.org/draft-04/schema#',
+  type: 'object',
+  properties: {
+    type: {
+      enum: ['cross product']
+    },
+    items: {
+      type: 'array',
+      items: {
```
julien-f: Which fields are required?

src/schemas/job/crossProduct.js
julien-f: More descriptions would be great if not to difficult to write.

src/schemas/job/params-vector.js
```diff
@@ -0,0 +1,66 @@
...
+    },
+    minItems: 1
+  },
+  required: [
+    'type',
+    'items'
+  ]
+}
+
+/* Example:
```
julien-f: Use the simple but more complete example on Fabrice's gist.

src/schemas/job.js
```diff
@@ -0,0 +1,35 @@
...
+    key: {
+      type: 'string'
+      // TODO description
+    },
+    method: {
+      type: 'string',
+      description: 'called method'
+    },
+    paramsVector: {
+      '$ref': 'job/crossProduct.js'
```
julien-f: JS import instead.

src/schemas/job.js
```diff
@@ -0,0 +1,35 @@
...
+    type: {
+      enum: ['call']
+    },
+    id: {
+      type: 'string',
+      description: 'job identifier'
+    },
+    userId: {
+      type: 'string',
+      description: 'identifier of the user who have created the job (the grants of the user are used by the job)'
```
julien-f: grants → priviledges/permissions/capabilities.

src/schemas/job/params-vector.js
```diff
@@ -0,0 +1,59 @@
+export default {
+  $schema: 'http://json-schema.org/draft-04/schema#',
+  type: 'object',
+  properties: {
+    type: {
+      enum: ['cross product']
```
marsaud: crossProduct

src/schemas/job/params-vector.js
```diff
@@ -0,0 +1,59 @@
...
+          }
+        },
+        required: [
+          'type',
+          'values'
+        ]
+      },
+      minItems: 1
+    },
+    minItems: 1
```
marsaud: items here refers to a property of an object, not the items of an array type, so I don't think this has sense

src/schemas/job.js
```diff
@@ -0,0 +1,35 @@
...
+    paramsVector
+  },
+  required: [
+    'type',
+    'id',
+    'userId',
+    'method',
+    'paramsVector'
+  ]
+}
```
marsaud: I added recently a not required "name" string property (human readable name)

src/schemas/job.js
```diff
@@ -0,0 +1,35 @@
...
+      description: 'called method'
+    },
+    paramsVector
+  },
+  required: [
+    'type',
+    'id',
+    'userId',
+    'method',
+    'paramsVector'
```
marsaud: paramsVector is not required, some methods have no params. If you have found code that would crash with an undefined paramsVector, it may happen I'm fixing this in #130 
We can wait for the merge of my PR to change this

src/schemas/job.js
```diff
@@ -0,0 +1,35 @@
...
+      type: 'string'
+      // TODO description
+    },
+    method: {
+      type: 'string',
+      description: 'called method'
+    },
+    paramsVector
+  },
+  required: [
```
marsaud: key must be required

julien-f: Why? What is it? Could you provide a short description for it?

src/schemas/job.js
```diff
@@ -0,0 +1,39 @@
...
+    type: {
+      enum: ['call']
+    },
+    id: {
+      type: 'string',
+      description: 'job identifier'
+    },
+    name: {
+      type: 'string',
+      description: 'human readable name'
```
julien-f: I think the description could be better…
@marsaud What is it exactly?

marsaud: A human readable name to identify the job in alist, better than 'job:417'...an explicit name for humans

src/schemas/job.js
marsaud: The key is a link to a special xo-web feature that is managed through a dedicated UI. Like "rollingSnapshot" for example. It is used to dispatch the existing jobs and schedules in the right screens for visualisation and edition, so it is fully required. A new key like "genericJob" will be introduced by xo-web#493.

Description could be "related xo-web feature" ?...

PR #139, opened 2015-11-20
Save enabled state for each server

Wescoeur:


src/xo.js
```diff
@@ -960,14 +962,21 @@ export default class Xo extends EventEmitter {
...
-    const server = await this._getXenServer(id)
-
+  async _updateXenServer (server, {host, username, password, enabled}) {
     if (host) server.set('host', host)
     if (username) server.set('username', username)
     if (password) server.set('password', password)
 
-    await this._servers.update(server)
+    if (enabled !== undefined) {
+      server.enabled = enabled ? true : undefined
```
julien-f: Incorrect, you can see for the other attributes that they are not plain properties.

src/xo.js
```diff
@@ -960,14 +962,21 @@ export default class Xo extends EventEmitter {
     }
   }
 
-  async updateXenServer (id, {host, username, password}) {
-    const server = await this._getXenServer(id)
-
+  async _updateXenServer (server, {host, username, password, enabled}) {
```
julien-f: If no longer used, you can merge it back in `updateXenServer()`.

src/xo.js
```diff
@@ -1081,6 +1088,7 @@ export default class Xo extends EventEmitter {
     }
 
     xapi.xo.uninstall()
+
```
julien-f: Not necessary, can you remove it?

PR #146, opened 2015-11-26
Start VM export only when necessary.

Wescoeur:


julien-f:
Can you change you commit message?

PR #148, opened 2015-11-27
The plugins configurations can be cleaned.

Wescoeur:


src/xo.js
```diff
@@ -1547,6 +1547,16 @@ export default class Xo extends EventEmitter {
     plugin.loaded = false
   }
 
+  async purgePluginConfiguration (id) {
+    const plugin = this._getRawPlugin(id)
+
+    if (!plugin.configurationSchema) {
```
julien-f: Remove this condition which is unnecessary and can cause issues.

src/api/plugin.js
```diff
@@ -86,3 +86,19 @@ unload.params = {
 }
 
 unload.permission = 'admin'
+
+// -------------------------------------------------------------------
+
+export async function purge ({ id }) {
```
julien-f: Rename to `purgeConfiguration`.

PR #155, opened 2015-12-04
Installed plugins are detected even if the config file does not show …

pdonias:
…their configurations. Thus, their configurations are empty.


julien-f:
The commit message is incorrect: the plugins are now never discovered from the config file.

PR #159, opened 2015-12-07
Smb remote

marsaud:


src/remote-handler.js
```diff
@@ -1,17 +1,82 @@
...
 import fs from 'fs-promise'
+import makeError from 'make-error'
+import Smb2 from 'smb2'
 import {exec} from 'child_process'
-
-import {
-  forEach,
-  promisify
-} from './utils'
+import {forEach, promisify} from './utils'
```
julien-f: This change is not necessary.

src/remote-handler.js
```diff
@@ -1,17 +1,82 @@
...
+  _getClient (remote) {
+    if (!(remote.id in this._clients)) {
+      try {
+        const client = new Smb2({
+          share: `\\${remote.host}`,
+          domain: remote.domain,
+          username: remote.username,
+          password: remote.password
+        })
+        Bluebird.promisifyAll(client)
```
julien-f: You can use `promisifyAll()` from utils.

src/remote-handler.js
```diff
@@ -1,17 +1,82 @@
...
+        this._clients[remote.id] = client
+      } catch (error) {
+        delete this._clients[remote.id]
+        throw new SmbClientFailure(error.message)
```
julien-f: Why wrapping the error? It removes the original stacktrace…

src/remote-handler.js
```diff
@@ -1,17 +1,82 @@
...
+        const client = new Smb2({
+          share: `\\${remote.host}`,
+          domain: remote.domain,
+          username: remote.username,
+          password: remote.password
+        })
+        Bluebird.promisifyAll(client)
+        this._clients[remote.id] = client
+      } catch (error) {
+        delete this._clients[remote.id]
```
julien-f: I don't believe this line to be necessary.

src/remote-handler.js
```diff
@@ -1,17 +1,82 @@
...
+
+  async forget (remote) {
+    if (remote.id in this._clients) {
+      this._clients[remote.id] && this._clients[remote.id].close && this._clients[remote.id].close()
+      delete this._clients[remote.id]
+    }
+  }
+
+  async list (remote) {
+    const enabled = remote.enabled
```
julien-f: Perfect case to use destructuring: `const { enabled } = remote`

src/remote-handler.js
```diff
@@ -1,17 +1,82 @@
...
+    const files = await this._getClient(remote).readdirAsync(remote.share)
+    if (!enabled) { // Do not leave the remote enabled after a one shot read access
+      this.forget(remote)
+    }
+    return filter(files, file => endsWith(file, '.xva'))
+  }
+
+  createReadStream (remote, file) {
+    return this._getClient(remote).createReadStream(remote.path + '/' + file) // TODO Smb2.createReadStream does not exist
```
julien-f: Use ES6 template strings instead.

src/remote-handler.js
```diff
@@ -1,17 +1,82 @@
...
+  }
+
+  async disableAll (remotes) {
+    for (let remote of remotes) {
```
julien-f: `for ( of )` → `const`.

src/remote-handler.js
```diff
@@ -1,17 +1,82 @@
...
+      await this.forget(remote)
```
julien-f: Can it not be done in parallel?

src/remote-handler.js
```diff
@@ -1,17 +1,82 @@
...
+    }
+  }
+}
+
+class NfsHandler {
```
julien-f: You should split remote handlers in different files, e.g.: `remote-handlers/nfs.js`.

src/remote-handler.js
```diff
@@ -83,6 +148,15 @@ class NfsMounter {
     return mount
   }
 
+  async list (local) {
+    const files = await fs.readdir(local.path)
+    return filter(files, file => endsWith(file, '.xva'))
```
julien-f: Would not it be clearer to remove `files` and inline the `readdir()`? (It is a real question :smile:)

src/remote-handler.js
```diff
@@ -83,6 +148,15 @@ class NfsMounter {
...
+  }
+
+  createReadStream (mount, file) {
+    return fs.createReadStream(mount.path + '/' + file)
```
julien-f: ES6 template strings.

src/remote-handler.js
```diff
@@ -115,27 +189,53 @@ class LocalHandler {
...
     return local
   }
+
+  async list (local) {
+    const files = await fs.readdir(local.path)
+    return filter(files, file => endsWith(file, '.xva'))
+  }
+
+  createReadStream (local, file) {
+    return fs.createReadStream(local.path + '/' + file)
```
julien-f: ES6 template strings.

src/remote-handler.js
```diff
@@ -115,27 +189,53 @@ class LocalHandler {
     }
     return local
   }
+
+  async list (local) {
+    const files = await fs.readdir(local.path)
+    return filter(files, file => endsWith(file, '.xva'))
```
julien-f: Same question about inlining `files`.

src/remote-handler.js
```diff
@@ -115,27 +189,53 @@ class LocalHandler {
...
 
 export default class RemoteHandler {
   constructor () {
     this.handlers = {
-      nfs: new NfsMounter(),
-      local: new LocalHandler()
+      nfs: new NfsHandler(),
+      local: new LocalHandler(),
+      smb: new SmbHandler()
     }
```
julien-f: If all handlers are instantiated here, I suppose they are stateless?

src/remote-handler.js
```diff
@@ -115,27 +189,53 @@ class LocalHandler {
...
+  }
+
+  createReadStream (remote, file) {
+    return this._getHandler(remote).createReadStream(remote, file)
   }
 
   async disableAll (remotes) {
     const promises = []
-    forEach(['local', 'nfs'], type => promises.push(this.handlers[type].disableAll(filter(remotes, remote => remote.type === type))))
+    forEach(['local', 'nfs', 'smb'], type => promises.push(this.handlers[type].disableAll(filter(remotes, remote => remote.type === type))))
```
julien-f: Could you not iterate directly on `this.handlers`?
It would avoid repeating the handler types everywhere…

src/remote-handler.js
marsaud: True. I don't like to see trivial code lines taking the room of 4 on a screen or diff, but if you don't wanna change...

src/remote-handler.js
marsaud: I don't catch your idea...

marsaud: I do. Well I agree it could have been written some other way...

julien-f: It was a question. If you think it could be better written, feel free to leave a comment :)

src/remote-handlers/nfs.js
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  _matchesRealMount (mount) {
+    return this._fullPath(mount.path) in this._realMounts
+  }
+
+  async _mount (mount) {
+    const path = this._fullPath(mount.path)
+    await fs.ensureDir(path)
+    return await execAsync(`mount -t nfs ${mount.host}:${mount.share} ${path}`)
```
julien-f: You should use `fs.execFile()` or [execa](https://github.com/sindresorhus/execa) to avoid issues (e.g. variables containing special chars).

src/remote-handlers/smb.js
```diff
@@ -0,0 +1,63 @@
+import endsWith from 'lodash.endswith'
+import filter from 'lodash.filter'
+import Smb2 from 'smb2'
+import {promisifyAll} from './utils'
+
+export default class SmbHandler {
+  constructor () {
+    this._clients = {}
```
julien-f: IMHO, it would be much simpler and more efficient to simply have an instance `_handler` property on each remote.

src/remote-handlers/smb.js
```diff
@@ -0,0 +1,63 @@
...
+        remote.enabled = false
+        remote.error = error.message
+      }
+    } else {
+      this.forget(remote)
+    }
+  }
+
+  async forget (remote) {
+    if (remote.id in this._clients) {
```
julien-f: `const { id } = remote` would simplify your code.

src/remote-handlers/smb.js
```diff
@@ -0,0 +1,63 @@
...
+      this._clients[remote.id] && this._clients[remote.id].close && this._clients[remote.id].close()
+      delete this._clients[remote.id]
+    }
+  }
+
+  async list (remote) {
+    const {enabled} = remote
+    const files = await this._getClient(remote).readdirAsync(remote.share)
+    if (!enabled) { // Do not leave the remote enabled after a one shot read access
```
julien-f: Is this the correct condition?

src/remote-handlers/nfs.js
marsaud: fs.execFile ?...

julien-f: My bad, `child_process.execFile()` :p

src/remote-handlers/smb.js
marsaud: This may change after maturation, for the moment, only "enabled" remotes maintain an instance of smb2 client opened, so when I instatiate and open one just for listing, I forget it right after.
For recall, a "disabled" remote expresses the will to NOT use it for backup right now... I think we can still let the possibility to explore it.

julien-f: Ok, thanks for the explanation.

src/remote-handlers/local.js
```diff
@@ -0,0 +1,33 @@
...
+      }
+    }
+    return local
+  }
+
+  async list (local) {
+    return filter(await fs.readdir(local.path), file => endsWith(file, '.xva'))
+  }
+
+  getFsMethod () {
```
julien-f: Why not exposing directly high level methods which handle paths relatively to the remote and ensure dirs existence when required?
- `createReadStream(path)`
- `createWriteStream(path)`

`readdir()` is not necessary because you already have `list()`.

julien-f:
I believe the global architecture should be:
- a *remote* is an object with various metadata
- a *handler* is an instance of the abstract class `RemoteHandlerAbstract` which provides default implementation for public methods
    - do nothing for optional methods (`noop`)
    - throw a *not implemented* error for mandatory methods
- `xo.getRemoteHandler(remote)` returns an instance of the specific handler for a given remote

What's your opinion?

julien-f:
If remotes are plain objects and instance of a class, I would like the schema of a remote to be documented in `schemas/remote.js`.

src/remote-handlers/local.js
julien-f: Exposing specific methods gives a better control of what is used from the outside.

marsaud:
Why not, my "remote-handler" as a component seems less elegant than your RemoteHandlerAbstract class hierarchy.

For the remote objetcs, I've put up something a bit strange, as the model/remote considers mainly a url property, and there is this xo._developRemote that makes a kind of parse to ease work with the remote... if I create a schema it would probably describe what can pop out of the "_develop" ?...

julien-f:
I don't think it's necessary to go into implementation details, I just want to be able to quickly know what are the fields that are always present on a remote and which can be used in public code.

marsaud:
Instead of a `xo.getRemoteHandler` I suggest that `xo.getRemote` returns a remote object with it's own handler.

This :
```javascript
let remote = await xo.getRemote(id)
let handler = xo.getRemoteHandler(remote)
return await handler.list(remote)
```
will turn to that :
```javascript
let remote = await xo.getRemote()
return await remote.list()
```

src/api/vm.coffee
```diff
@@ -493,6 +493,13 @@ exports.snapshot = snapshot
 #---------------------------------------------------------------------
 
 rollingDeltaBackup = $coroutine ({vm, remote, tag, depth}) ->
+  _remote = yield @getRemote remote
+  if not _remote?.path?
+    throw new Error "No such Remote #{remote}"
+  if not _remote.enabled
+    throw new Error "Backup remote #{remote} is disabled"
+  if _remote.type == 'smb'
+    throw new Error "Delta Backup is not supported for smb remotes"
```
julien-f: Should not all of this be checked in `Xo#rollingDeltaVmBackup()`?

marsaud: Why and why not...as you want.

julien-f: Why not?
Because these tests are not related to the API and should be done each time `rollingDeltaVmBackup()` is called.

marsaud: Yeah, you're right ^^

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,49 @@
...
+  async sync () {
+    throw new Error('Not implemented')
+  }
+
+  async forget () {
+    throw new Error('Not implemented')
+  }
+
+  async outputFile (file, data, options) {
+    throw new Error('Not implemented')
```
julien-f: The default implementation could be based on `this.createReadStream()`.

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  async readFile (file, options) {
+    throw new Error('Not implemented')
```
julien-f: Similar: could be based on `createReadStream()`.

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  async rename (oldPath, newPath) {
+    throw new Error('Not implemented')
+  }
+
+  async list (dir = undefined) {
```
julien-f: What does mean the value `undefined`? `/`?

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,49 @@
+export default class RemoteHandlerAbstract {
+  constructor (remote) {
+    this._remote = remote
+  }
+
+  set (remote) {
```
julien-f: I do believe that there should be one handler instance by remote.

Because with the current pattern, I could use a remote handler and have the linked remote change without even knowing it:
```js
const handler = new RemoteHandlerXX()

async function listRemote (remote) {
  handler.set(remote)
  console.log(await handler.list())
}

await Promise.all([
  listRemote(new Remote('foo'))
  listRemote(new Remote('bar'))
])
```

src/remote-handlers/abstract.js
marsaud: What do you mean, using a stream by default to output files ?

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,49 @@
...
+    this._remote = remote
+  }
+
+  async sync () {
```
julien-f: A common pattern is to have all public methods call private methods which are meant to be defined in inherited classes:

```js
sync () {
  return this._sync()
}
_sync () {
  throw new Error('not implemented')
}
```

It is handy because it allows us to add some common code around the specific implementations (e.g. [collection.js](/vatesfr/xo-server/blob/next-release/src/collection.js)).

src/remote-handlers/abstract.js
marsaud: Yes

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,49 @@
...
+
+  async rename (oldPath, newPath) {
+    throw new Error('Not implemented')
+  }
+
+  async list (dir = undefined) {
+    throw new Error('Not implemented')
+  }
+
+  async createReadStream (file) {
```
julien-f: I would like very much for this method to automatically call `this.getSize()` and add a `length` property to the stream if possible.

src/remote-handlers/abstract.js
julien-f: Why not setting '/' as the default value then?

src/remote-handlers/abstract.js
julien-f: Not at all, just allowing the implementation class to not implement this method and have it default to:

```js
const stream = this.createOutputStream(file)
const promise = eventToPromise(stream, 'finish')
stream.end(data)
return promise
```

src/remote-handlers/nfs.js
```diff
@@ -0,0 +1,119 @@
+import fs from 'fs-promise'
+import RemoteHandlerAbstract from './abstract'
+import {dirname} from 'path'
+import {exec} from 'child_process'
+import {forEach, promisify} from '../utils'
+
+const execAsync = promisify(exec)
```
julien-f: You can use [execa](https://github.com/sindresorhus/execa) instead (already used in xo-server).

src/remote-handlers/local.js
```diff
@@ -0,0 +1,70 @@
...
+    super(remote)
+    this.forget = noop
+  }
+
+  _getFilePath (file) {
+    const parts = [this._remote.path]
+    if (file) {
+      parts.push(file)
+    }
+    return parts.join('/')
```
julien-f: Use [`path.join.apply(null, parts)`](http://devdocs.io/node_lts/path#path_path_join_path1_path2) instead.

julien-f: Better still, use [`path.resolve()`](http://devdocs.io/node_lts/path#path_path_resolve_from_to) and check that the root path is still the prefix.

src/remote-handlers/abstract.js
marsaud: I'm not sure. The parameter is a path relative to the remote "place".

If we want to list a specific folder of the remote, we would pass "foo/bar", not "/foo/bar", so why pass "/" if we want the top folder ?

julien-f: Because this is the absolute path for this remote, just like when you load `/index.html` from `google.com` the server reads the file relatively to the web root.

Anyway, if you prefer relative paths, you can use `'.'` as the default value.

src/remote-handlers/nfs.js
```diff
@@ -0,0 +1,119 @@
...
+
+export default class NfsHandler extends RemoteHandlerAbstract {
```
julien-f: It could inherit from `LocalHandler`.

src/api/vm.coffee
```diff
@@ -572,12 +572,18 @@ exports.rollingSnapshot = rollingSnapshot
...
 
-backup = $coroutine ({vm, pathToFile, compress, onlyMetadata}) ->
-  yield @backupVm({vm, pathToFile, compress, onlyMetadata})
+backup = $coroutine ({vm, remoteId, file, compress, onlyMetadata}) ->
+  remote = yield @getRemote remoteId
+  if not remote?.path?
+    throw new Error "No such Remote #{remoteId}"
+  if not remote.enabled
+    throw new Error "Backup remote #{remoteId} is disabled"
+  yield @backupVm({vm, remoteId, file, compress, onlyMetadata})
```
marsaud: I didn't see that on the last diff, I'll push down the errors in the lower layer

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,101 @@
...
+
+  async readFile (file, options) {
+    return await this._readFile(file, options)
+  }
+
+  async _readFile (file, options) {
+    const stream = this.createReadStream(file, options)
+    let data = ''
+    stream.on('data', d => data += d)
+    await eventToPromise(stream, 'end')
```
julien-f: You can use [get-stream](https://github.com/sindresorhus/get-stream) which is already a dependency.

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,101 @@
...
+  constructor (remote) {
+    this._remote = this._getInfo({...remote})
+  }
+
+  _getInfo (remote) {
+    throw new Error('Not implemented')
+  }
+
+  async sync () {
+    return await this._sync()
```
julien-f: `return` + `await` is unnecessary (`async` is because it guarantees that a promise will be returned) and can have a perf impact.

It might be useful for doc purposes but not for such a simple function.

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,101 @@
...
+
+  async _rename (oldPath, newPath) {
+    throw new Error('Not implemented')
+  }
+
+  async list (dir = '.') {
+    return await this._list(dir)
+  }
+
+  async _list (dir = '.') {
```
julien-f: Do not provide default values in private implementations.

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,101 @@
...
+    throw new Error('Not implemented')
+  }
+
+  async createReadStream (file, options) {
+    const length = await this.getSize(file)
+    const stream = await this._createReadStream(file)
```
julien-f: Use parallelism.

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,101 @@
...
+  async list (dir = '.') {
+    return await this._list(dir)
+  }
+
+  async _list (dir = '.') {
+    throw new Error('Not implemented')
+  }
+
+  async createReadStream (file, options) {
+    const length = await this.getSize(file)
```
julien-f: Move after `this._createReadStream()`, and only do it if necessary (missing property, not different to `null`) because some implementations (e.g. HTTP for instance) may be able to provide it directly.

If this method fails, simply do not add the `length` property: some implementation may not be able to provide such an information.

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,101 @@
...
+  async _unlink (file) {
+    throw new Error('Not implemented')
+  }
+
+  async getSize (file) {
+    return await this._getSize(file)
+  }
+
+  async _getSize (file) {
+    throw new Error('Not implement')
```
julien-f: `'Not implemented'` :stuck_out_tongue:.

src/remote-handlers/local.js
```diff
@@ -0,0 +1,84 @@
+import fs from 'fs-promise'
+import RemoteHandlerAbstract from './abstract'
+import startsWith from 'lodash.startswith'
+import {noop} from '../utils'
```
julien-f: Put spaces inside braces: this might be enforced by standard one day.

src/remote-handlers/local.js
```diff
@@ -0,0 +1,84 @@
...
+import {dirname, resolve} from 'path'
+
+export default class LocalHandler extends RemoteHandlerAbstract {
+  _getInfo (remote) {
+    if (!startsWith(remote.url, 'file://')) {
+      throw new Error('Incorrect remote type')
+    }
+    this.type = 'local'
```
julien-f: Move into a `get type () { return 'local' }` directly in the class.

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,103 @@
...
+    stream.end(data)
+    return promise
+  }
+
+  async readFile (file, options) {
+    return this._readFile(file, options)
+  }
+
+  async _readFile (file, options) {
+    const stream = await this.createReadStream(file, options)
```
julien-f: Why intermediary variable?

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,103 @@
...
+
+  async _list (dir) {
+    throw new Error('Not implemented')
+  }
+
+  async createReadStream (file, options) {
+    const stream = await this._createReadStream(file)
+    if (!('length' in stream) || stream.length === null) {
+      try {
+        const length = await this.getSize(file)
```
julien-f: Why this intermediary variable?

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,103 @@
+import eventToPromise from 'event-to-promise'
+import getStream from 'get-stream'
+
+export default class RemoteHandlerAbstract {
+  constructor (remote) {
+    this._remote = this._getInfo({...remote})
```
julien-f: Why this `_getInfo()`?
Just let every implementation create is own constructor and manage its own state, no?

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  _getInfo (remote) {
+    throw new Error('Not implemented')
+  }
+
+  async sync () {
```
julien-f: Please document this method as it is not clear what it does when it is called…

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,103 @@
...
+    return this._sync()
+  }
+
+  async _sync () {
+    throw new Error('Not implemented')
+  }
+
+  async forget () {
```
julien-f: Same as `sync()`.

src/remote-handlers/nfs.js
```diff
@@ -0,0 +1,88 @@
+import fs from 'fs-promise'
+import LocalHandler from './local'
+import startsWith from 'lodash.startswith'
+import execa from 'execa'
+import { forEach } from '../utils'
```
julien-f: I usually split my imports in two groups *external dependencies* and *local dependencies*.
In each of these groups, imports are sorted: first whether they are default or named imports then alphabetically.

I really should write a coding style guide.

src/remote-handlers/nfs.js
```diff
@@ -0,0 +1,88 @@
...
+
+  async _loadRealMounts () {
+    let stdout
+    try {
+      [stdout] = await execa('findmnt', ['-P', '-t', 'nfs,nfs4', '--output', 'SOURCE,TARGET', '--noheadings'])
+    } catch (exc) {
+      // When no mounts are found, the call pretends to fail...
+    }
+    const mounted = {}
+    if (stdout) {
```
julien-f: This `if` could be removed and the block moved directly inside the `try` no?

src/remote-handlers/abstract.js
marsaud: I wish to keep this ._remote property, to isolate data taken from the remote, and others properties  specific to the handler's needs. It could happen that ._remote would be "empty" in some kind of remote, it doesn't appear as a problem to me.

src/remote-handlers/nfs.js
marsaud: Ok for external/local. Then, in each group you should expect simple alpha order as provided, for example by sublime's F9. Otherwise, it appears to me too tricky and difficult to spread through a team, for a poor gain.

julien-f: Sounds good to me.

marsaud: This level of detail will lead to many changes/diffs on the import section, and we will so lose the benefit to standardize an ordering rule.

marsaud: Just a somehow related point, when breaking multiple named imports on several lines, automatic alpha ordering is not usable. Not very important.

src/api/remote.js
```diff
@@ -15,7 +15,8 @@ get.params = {
   id: {type: 'string'}
 }
 
-export async function list (id) {
+export async function list ({id}) {
+  console.log(id)
```
julien-f: Avoid all console.log() in code!

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,118 @@
...
+import getStream from 'get-stream'
+
+import {
+  parse
+} from 'xo-remote-parser'
+
+export default class RemoteHandlerAbstract {
+  constructor (remote) {
+    this._remote = parse({...remote})
+    if (!this._remote.type === this.type) {
```
julien-f: Maybe `!==`?

src/remote-handlers/abstract.js
```diff
@@ -0,0 +1,118 @@
...
+      throw new Error('Incorrect remote type')
+    }
+  }
+
+  _getInfo (remote) {
```
julien-f: I don't understand, this method is no longer used but still specified?

src/remote-handlers/local.js
```diff
@@ -0,0 +1,84 @@
...
+    return 'local'
+  }
+
+  _getFilePath (file) {
+    const parts = [this._remote.path]
+    if (file) {
+      parts.push(file)
+    }
+    const path = resolve.apply(null, parts)
+    if (!startsWith(path, this._remote.path)) {
```
julien-f: Just for info: the main goal of this test is to avoid escaping the root path (e.g. `../../etc/ssl/toto.crt`).

src/xo-mixins/remotes.js
```diff
@@ -110,9 +109,11 @@ export default class {
   }
 
   // TODO: Should it be private?
-  async disableAllRemotes () {
+  async forgetAllRemotes () {
     const remotes = await this.getAllRemotes()
-    this.remoteHandler.disableAll(remotes)
+    for (let remote of remotes) {
+      await remote.handler.forget()
```
julien-f: Is it still correct?

PR #161, opened 2015-12-10
Set default sr

olivierlambert:
Fix https://github.com/vatesfr/xo-web/issues/572

### DIFF
--- a/src/api/pool.js
+++ b/src/api/pool.js
@@ -49,8 +49,8 @@ setDefaultSr.params = {
 }
 
 setDefaultSr.resolve = {
-  pool: ['pool', 'pool', 'administrate']
-  sr: ['sr', 'sr']
+  pool: ['pool', 'pool', 'administrate'],
+  sr: ['sr', 'SR']
 }
 // -------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


src/api/pool.js
```diff
@@ -35,6 +35,25 @@ set.resolve = {
...
+    type: 'string'
+  },
+  sr: {
+    type: 'string'
+  }
+}
+
+setDefaultSr.resolve = {
+  pool: ['pool', 'pool', 'administrate']
+  sr: ['sr', 'sr']
```
julien-f: [Unexpected token](https://travis-ci.org/vatesfr/xo-server/jobs/95994547#L1015).

src/xapi.js
```diff
@@ -317,6 +317,12 @@ export default class Xapi extends XapiBase {
 
   // =================================================================
 
+  async setDefaultSr (srId) {
+    await this.call('pool.set_default_SR', this.getObject(srId).$ref)
```
julien-f: Missing arg: pool reference.

You should do this instead:
```js
this._setObjectProperties(this.pool, {
  default_SR: this.getObject(srId).$ref
})
```

PR #170, opened 2015-12-14
[WIP] Connection to a server with a read-only status.

pdonias:


src/xo.js
```diff
@@ -968,11 +968,17 @@ export default class Xo extends EventEmitter {
...
+  async registerXenServer ({host, username, password, readOnly}) {
     // FIXME: We are storing passwords which is bad!
     //        Could we use tokens instead?
     // TODO: use plain objects
-    const server = await this._servers.create({host, username, password, enabled: 'true'})
+    const server = await this._servers.create({
+      host,
+      username,
+      password,
+      readOnly,
```
julien-f: You cannot pass a boolean directly, they are not correctly handled by the database (see `updateXenServer()`).

src/api/server.js
```diff
@@ -131,3 +132,22 @@ disconnect.params = {
     type: 'string'
   }
 }
+
+// -------------------------------------------------------------------
+
+export async function setReadOnly ({id, readOnly}) {
+  await this.setReadOnly({id, readOnly})
+}
```
julien-f: Not necessary: remove.

src/xo.js
```diff
@@ -968,11 +968,18 @@ export default class Xo extends EventEmitter {
 
   // -----------------------------------------------------------------
 
-  async registerXenServer ({host, username, password}) {
+  async registerXenServer ({host, username, password, readOnly}) {
     // FIXME: We are storing passwords which is bad!
     //        Could we use tokens instead?
     // TODO: use plain objects
-    const server = await this._servers.create({host, username, password, enabled: 'true'})
+    const ro = readOnly ? 'true' : undefined
```
julien-f: It could be inline.

src/xo.js
```diff
@@ -1055,7 +1066,8 @@ export default class Xo extends EventEmitter {
       auth: {
         user: server.username,
         password: server.password
-      }
+      },
+      readOnly: server.readOnly
```
julien-f: You should pass a proper boolean.

src/xo.js
```diff
@@ -968,11 +968,17 @@ export default class Xo extends EventEmitter {
 
   // -----------------------------------------------------------------
 
-  async registerXenServer ({host, username, password}) {
+  async registerXenServer ({host, username, password, readOnly}) {
```
julien-f: Explicit the default value for `readOnly`.

PR #171, opened 2015-12-15
allow edition of existing disks during VM creation

olivierlambert:


src/xapi.js
```diff
@@ -900,7 +901,20 @@ export default class Xapi extends XapiBase {
       )))
     }
 
-    // TODO: Create Cloud config drives.
+    // If orignal template had existing disks
+    //  AND they were modified (name, description) -> set new values
+    await Promise.all(mapToArray(existingDisks, async ({ size, srId, ...properties }, userdevice) => {
+      const vbd = find(vm.$VBDs, { userdevice })
```
julien-f: You must assert existence (<> `undefined`) before using it.

src/xapi.js
```diff
@@ -900,7 +901,20 @@ export default class Xapi extends XapiBase {
...
+      await this._setObjectProperties(vbd.$vdi, properties)
```
julien-f: Use a variable to avoid always accessing the property `vbd.$vdi`.

src/xapi.js
```diff
@@ -900,7 +901,20 @@ export default class Xapi extends XapiBase {
...
+      // if the disk is bigger
+      if (size > vbd.$vdi.size) {
+        await this.resizeVdi(vbd.$vdi, size)
```
julien-f: I don't think `resizeVdi()` expects an object.

PR #173, opened 2015-12-16
Possibility to edit the `vga` and `videoram` platform params of a VM

julien-f:


src/api/vm.coffee
```diff
@@ -320,6 +320,14 @@ set = $coroutine (params) ->
     else
       yield xapi.call 'VM.remove_from_other_config', ref, 'auto_poweron'
 
+  if 'vga' of params
+    {vga} = params
+    yield xapi.call 'VM.set_platform', ref, {vga}
```
julien-f: You cannot use `set_platform()` to update only a single entry because it will erase all others.

You should instead create a `Xapi#setVmProperties()` method similar to ([`Xapi#setPoolProperties()`](https://github.com/vatesfr/xo-server/blob/be19ad5f2ab7596cb3f34ba9a720f25e6395a22a/src/xapi.js#L304-L320)) which uses [`Xapi#_updateObjectMapProperty()`](https://github.com/vatesfr/xo-server/blob/be19ad5f2ab7596cb3f34ba9a720f25e6395a22a/src/xapi.js#L272-L292).

src/api/vm.coffee
```diff
@@ -320,6 +320,14 @@ set = $coroutine (params) ->
...
+
+  if 'videoram' of params
+    {videoram} = params
+    yield xapi.call 'VM.set_platform', ref, {videoram}
```
julien-f: Idem.

PR #177, opened 2015-12-18
The vm import call use a sr instead of a host.

Wescoeur:


src/api/vm.coffee
```diff
@@ -577,6 +577,27 @@ exports.backup = backup
 
 #---------------------------------------------------------------------
 
+importBackup = $coroutine ({remote, file, sr}) ->
+  yield @importVmFromRemote(remote, file, sr)
```
julien-f: Missing empty `return`.

src/xo.js
```diff
@@ -827,8 +827,8 @@ export default class Xo extends EventEmitter {
     }
   }
 
-  async importVmFromRemote (id, file, host) {
-    const remote = await this.getRemote(id)
+  async importVmFromRemote (remoteId, file, sr) {
```
julien-f: Rename to `importVmBackup()` for homogeneity.

PR #179, opened 2015-12-18
Handle VDI import/export.

julien-f:


julien-f:
Support of RAW format for export would be nice.

PR #182, opened 2015-12-22
Old vdi backups are properly removed.

Wescoeur:


julien-f:
Create an issue describing what this PR fixes.

Change title to: *Only delete VDI exports when VM backup is successful (fix vatesfr/xo-web#<n>)*.

PR #184, opened 2015-12-22
Disable vm start during delta import and explicit notification.

Wescoeur:


src/api-errors.js
```diff
@@ -50,3 +50,11 @@ export class AlreadyAuthenticated extends JsonRpcError {
     super('already authenticated', 4)
   }
 }
+
+// -------------------------------------------------------------------
+
+export class OperationBlocked extends JsonRpcError {
+  constructor (operation, reason) {
+    super(`${operation} blocked. (reason: ${reason})`, 5)
```
julien-f: I would prefer to use string constant for error code.

src/xapi.js
```diff
@@ -1284,7 +1287,15 @@ export default class Xapi extends XapiBase {
   }
 
   async startVm (vmId) {
-    await this._startVm(this.getObject(vmId))
+    try {
```
julien-f: Move this wrapping in the API.

src/xo.js
```diff
@@ -1147,6 +1147,9 @@ export default class Xo extends EventEmitter {
     // Because XenServer creates Vbds linked to the vdis of the backup vm if it exists.
     await xapi.destroyVbdsFromVm(vm.uuid)
 
+    // Disable start.
+    await xapi.addBlockedOperationToVm(vm.$id, 'start', 'Delta backups import...')
```
julien-f: Prefix reason with *[XO]*.

julien-f: Group with the name label change.

src/xo.js
```diff
@@ -1171,7 +1174,12 @@ export default class Xo extends EventEmitter {
         }
       )
     )
+
     await xapi._setObjectProperties(vm, { name_label: vmName })
+
+    // Reenable start.
+    await xapi.removeBlockedOperationFromVm(vm.$id, 'start')
```
julien-f: Group with name label restoration.

src/xo.js
```diff
@@ -1143,12 +1143,15 @@ export default class Xo extends EventEmitter {
...
     const vmName = vm.name_label
-    await xapi._setObjectProperties(vm, { name_label: `[Importing...] ${vmName}` })
 
     // Destroy vbds if necessary. Why ?
     // Because XenServer creates Vbds linked to the vdis of the backup vm if it exists.
     await xapi.destroyVbdsFromVm(vm.uuid)
 
+    // Disable start and change the VM name label during import.
+    await xapi.addForbiddenOperationToVm(vm.$id, 'start', 'Delta backup import...')
+    await xapi._setObjectProperties(vm, { name_label: `[Importing...] ${vmName}` })
```
julien-f: These should be done in parallel and should be moved just after the VM creation.

src/xo.js
```diff
@@ -1173,7 +1176,11 @@ export default class Xo extends EventEmitter {
         }
       )
     )
+
+    // Import done, reenable start and set real vm name.
+    await xapi.removeForbiddenOperationFromVm(vm.$id, 'start')
     await xapi._setObjectProperties(vm, { name_label: vmName })
```
julien-f: In parallel.

PR #186, opened 2015-12-30
add 'install_repository' support for vatesfr/xo-web/issues/627

olivierlambert:


src/xapi-object-to-xo.js
```diff
@@ -300,6 +300,7 @@ const TRANSFORMS = {
 
           return disks
         })(),
+        install_repository: otherConfig['install-repository'],
```
julien-f: Keep the properties ordered.

PR #211, opened 2016-01-28
Marsaudf clear logs#661

marsaud:


src/loggers/leveldb.js
```diff
@@ -46,6 +47,19 @@ export default class LevelDbLogger {
     return highland(this._db.createReadStream())
       .filter(({value}) => value.namespace === this._namespace)
   }
+
+  del (id) {
+    if (!Array.isArray(id)) {
+      id = [id]
+    }
+    forEach(id, id => {
+      this._db.get(id, (err, value) => {
```
julien-f: Is it really necessary to `get()` ?

marsaud: To respect the fact that the logger is constructed as "namespaced" yes, otherwise not.

julien-f: My bad, LGTM.

PR #213, opened 2016-02-02
Fixed cpuWeight removal for default

marsaud:


src/xapi.js
```diff
@@ -98,16 +101,22 @@ const put = (stream, {
...
+  return value.indexOf('.') === -1
+    ? `${value}.0`
+    : value
+}
 const asInteger = value => String(value)
 
 const filterUndefineds = obj => pick(obj, value => value !== undefined)
 
+const prepareXapiParam = (param) => {
+  if (isFinite(param) && !isInteger(param)) { param = asFloat(param) }
```
julien-f: Why `isFinite()`?

src/xapi.js
```diff
@@ -98,16 +101,22 @@ const put = (stream, {
...
+  if (isInteger(param)) { param = asInteger(param) }
```
julien-f: Either use `else if` or return early.

src/xapi.js
```diff
@@ -3,8 +3,11 @@ import every from 'lodash.every'
 import fatfs from 'fatfs'
 import fatfsBuffer, { init as fatfsBufferInit } from './fatfs-buffer'
 import find from 'lodash.find'
+import isBoolean from 'lodash.isboolean'
 import includes from 'lodash.includes'
+import isFinite from 'lodash.isfinite'
```
julien-f: Remove (or at least comment) if unused.

src/xapi.js
```diff
@@ -108,6 +111,12 @@ const asInteger = value => String(value)
 
 const filterUndefineds = obj => pick(obj, value => value !== undefined)
 
+const prepareXapiParam = (param) => {
+  // if (isFinite(param) && !isInteger(param)) { param = asFloat(param) }
+  if (isInteger(param)) { param = asInteger(param) }
```
julien-f: Either `else if` or return early.

src/xapi.js
```diff
@@ -108,6 +111,11 @@ const asInteger = value => String(value)
 
 const filterUndefineds = obj => pick(obj, value => value !== undefined)
 
+const prepareXapiParam = (param) => {
```
julien-f: While you're at it, I think it would be nice to also support objects:

1. `filterUndefineds()` entries
2. map to `prepareXapiParam()`

julien-f: Also, do not use parentheses for single param arrow functions.

PR #215, opened 2016-02-02
Add a boolean to detect if a VDI is a snapshot or not

olivierlambert:
@julien-f if you find a better way to write this without breaking everything in `xo-web`, go ahead. Or merge this if it sounds good.


julien-f:
We are going another way.

PR #216, opened 2016-02-03
VDI-snapshot type

marsaud:


src/xapi-object-to-xo.js
```diff
@@ -388,14 +388,13 @@ const TRANSFORMS = {
...
 
-  // TODO: should we have a VDI-snapshot type like we have with VMs?
   vdi (obj) {
     if (!obj.managed) {
       return
     }
 
     return {
-      type: 'VDI',
+      type: obj.is_a_snapshot ? 'VDI-snapshot' : 'VDI',
```
julien-f: `snapshot_time` and `snapshot_of` properties are only meaningful on `VDI-snapshot` objects.

PR #229, opened 2016-02-04
Resource sets

julien-f:


src/api/resource-set.js
```diff
@@ -0,0 +1,56 @@
+export function create ({ name }) {
+  return this.createResourceSet(name)
```
marsaud: Could we handle optional arrays of subjects and objets, here ?

src/api/resource-set.js
```diff
@@ -0,0 +1,56 @@
...
+export function addObject ({ object }) {
+  return this.addObjectToResourceSet(object)
+}
+
+addObject.permission = 'admin'
+
+// -------------------------------------------------------------------
+
+export function removeObject ({ object }) {
+  return this.removeObjectFromResourceSet(object)
```
marsaud: An object, really, not an object id ?

marsaud: Why not an array of objects ?

src/api/resource-set.js
```diff
@@ -0,0 +1,56 @@
...
+}
+
+removeObject.permission = 'admin'
+
+// -------------------------------------------------------------------
+
+export function addSubject ({ subject }) {
+  return this.addSubjectToResourceSet(subject)
```
marsaud: Same questions than for objects

src/api/resource-set.js
```diff
@@ -0,0 +1,56 @@
...
+    ? this.getAllResourceSets()
+    : this.getResourceSet(id)
+}
+
+get.permission = 'admin'
+
+// -------------------------------------------------------------------
+
+export function addObject ({ object }) {
+  return this.addObjectToResourceSet(object)
```
marsaud: Missing `set` parameter everywhere, no ?

julien-f: Indeed :p

src/api/resource-set.js
julien-f: Sounds good to me.

src/api/resource-set.js
julien-f: Maybe keeping these simple methods for manual use and have a `set({ id, name, objects, subjects })` better suited for your needs?

src/api/resource-set.js
```diff
@@ -0,0 +1,56 @@
...
+  return this.deleteResourceSet(id)
+}
+export { delete_ as delete }
+
+delete_.permission = 'admin'
+
+// -------------------------------------------------------------------
+
+export function get ({ id }) {
+  return id == null
```
marsaud: A getAll would be nice too.

PR #238, opened 2016-02-16
Correct the redis protocol in sample config

bartekrutkowski:
The existing sample configuration file documents the Redis uri string with 'tcp://' prefix string, while xo-server actually expects it to be 'redis://' instead.


julien-f:
Thanks :)

PR #245, opened 2016-02-26
vm.set(): only admins can change cpuWeight if VM in a resource set

pdonias:


src/api/vm.coffee
```diff
@@ -478,6 +478,8 @@ set = $coroutine (params) ->
       yield xapi.call 'VM.remove_from_other_config', ref, 'auto_poweron'
 
   if 'cpuWeight' of params
+    if resourceSet?
+      throw new Unauthorized() unless this.user.permission is 'admin'
```
julien-f: Move `unless` condition in the `if` to simplify lecture and ease future transition to JS.

julien-f:
Change commit message to this PR's title.

PR #256, opened 2016-03-03
Xapi#startVmOnCd(): fix for HVM (fix vatesfr/xo-web#794).

julien-f:


olivierlambert:
Tested and working. Merging.

PR #259, opened 2016-03-03
Delta backups works with samba

Wescoeur:
Many fixes (linux paths, not found windows errors support...)


package.json
```diff
@@ -121,7 +121,6 @@
     "semver": "^5.1.0",
     "serve-static": "^1.9.2",
     "stack-chain": "^1.3.3",
-    "through2": "^2.0.0",
```
julien-f: FIXME.

PR #268, opened 2016-03-25
Throw an error if a vm is halted and its stats are requested.

Wescoeur:


src/xapi-stats.js
```diff
@@ -528,6 +528,11 @@ export default class XapiStats {
   async getVmPoints (xapi, vmId, granularity) {
     const vm = xapi.getObject(vmId)
     const host = vm.$resident_on
+
+    if (!host) {
+      throw new Error(`VM ${vmId} is halted.`)
```
julien-f: Or the host could not be found…

PR #274, opened 2016-04-06
api.vm: Set the min/max ram values.

Wescoeur:


julien-f:
`vm.set(memory)` should not change the dynamic min.

julien-f:
LGTM, let's merge it.

PR #305, opened 2016-05-10
vm.`backupCall(...)` uses `id` instead of `vm`

Wescoeur:


src/api.js
```diff
@@ -272,6 +272,12 @@ export default class Api {
 
     try {
       await checkPermission.call(context, method)
+
+      if (params && params.id === undefined) {
```
julien-f: A comment explaining why this is done would be nice :)

src/api.js
```diff
@@ -272,6 +272,20 @@ export default class Api {
...
       await checkPermission.call(context, method)
+
+      // API methods are in a namespace.
+      // Some methods use the namespace or an id parameter like:
+      //
+      // vm.detachPci vm=<string>
+      // vm.ejectCd id=<string>
+      //
+      // The goal here is to force each API call with an id/vm parameter(s)
+      // for simplify calls to the API.
```
julien-f: *for* → *to*.

src/api.js
```diff
@@ -272,6 +272,20 @@ export default class Api {
...
     try {
       await checkPermission.call(context, method)
+
+      // API methods are in a namespace.
+      // Some methods use the namespace or an id parameter like:
+      //
+      // vm.detachPci vm=<string>
+      // vm.ejectCd id=<string>
+      //
+      // The goal here is to force each API call with an id/vm parameter(s)
```
julien-f: *The goal here is to standardize the calls by always providing an `id` parameter when possible.*

Issue #309, opened 2016-05-12
XO inserts incorrect "other-config" value for pool auto_poweron


Fohdeesha:
on a fresh install of both Xenserver and XO, using the GUI to change a VM to auto-power-on does not work as intended. For this to work both the xenserver pool and vm must have the "other-config" value "auto_poweron=true". XO correctly adds this to the VM properties, however it adds to the pool config "auto_poweron=on" when it should be "auto_poweron=true".

This incorrect value can be seen when running "xe pool-param-get uuid=your-pool-uuid param-name=other-config" after using the GUI to enable the function. 

It returns "auto_poweron: on". Manually correcting this with a "xe pool-param-set uuid=your-pool-uuid other-config:auto_poweron=true" fixes the issue and VM's auto-boot when the host is rebooted as expected

source for correct pool value: http://support.citrix.com/article/CTX133910

olivierlambert:
Thanks for reporting this! Indeed you are right.

Fohdeesha:
Just found the location that needs changed, line 431, from 'on' to 'true' - I'd do a pull request and change it myself but I'm brand new to github and would probably break something :P

https://github.com/vatesfr/xo-server/blob/next-release/src/xapi.js#L431

olivierlambert:
You can try. See https://xen-orchestra.com/docs/contributing.html#pull-requests :)

Fohdeesha:
working on it as we speak!

Fohdeesha:
https://github.com/vatesfr/xo-server/pull/310

:) Thanks for the help!

PR #310, opened 2016-05-12
Fix pool autopoweron value from on to true

Fohdeesha:
when enabling auto power on for the pool, the other_config value should
be "true", not "on" - http://support.citrix.com/article/CTX133910

as discussed here - https://github.com/vatesfr/xo-server/issues/309


julien-f:
Thank you!

PR #316, opened 2016-05-20
fix(Xapi#_createVm()): license issue with Dundee

julien-f:


julien-f:
See vatesfr/xo-web#964.

PR #324, opened 2016-05-27
feat(sr): add disconnect and connect all pbds to a SR

olivierlambert:


src/api/sr.js
```diff
@@ -76,6 +76,34 @@ forget.resolve = {
 
 // -------------------------------------------------------------------
 
+export async function connectAllSrPbds ({SR}) {
```
julien-f: `connectAllPbds()`.

PR #329, opened 2016-06-06
updateJob protects userId from modification

marsaud:


src/xo-mixins/jobs.js
```diff
@@ -45,6 +45,8 @@ export default class {
   }
 
   async updateJob (job) {
```
julien-f: I would prefer to simply destructure the parameter to get only the updatable properties.

What do you think?

marsaud: Yeah, I'll write this between this evening and tomorrow.

PR #333, opened 2016-06-08
Remote handlers maintained for remote-parser 3

marsaud:


src/remote-handlers/abstract.js
```diff
@@ -1,5 +1,6 @@
 import eventToPromise from 'event-to-promise'
 import through2 from 'through2'
+import assign from 'lodash.assign'
```
julien-f: Sorted imports.

src/remote-handlers/abstract.js
```diff
@@ -15,7 +16,7 @@ import {
 
 export default class RemoteHandlerAbstract {
   constructor (remote) {
-    this._remote = parse({...remote})
+    this._remote = assign({...remote}, parse(remote.url))
```
julien-f: You can use `{ ...remote, ...parse(remote.url) }`, assign is not needed here.

src/remote-handlers/local.js
```diff
@@ -12,16 +12,20 @@ import {
 
 export default class LocalHandler extends RemoteHandlerAbstract {
   get type () {
-    return 'local'
+    return 'file'
```
julien-f: Does that make sense to you? I do not want to push a change you're uncomfortable with.

src/remote-handlers/local.js
```diff
@@ -30,8 +34,8 @@ export default class LocalHandler extends RemoteHandlerAbstract {
   async _sync () {
     if (this._remote.enabled) {
       try {
-        await fs.ensureDir(this._remote.path)
-        await fs.access(this._remote.path, fs.R_OK | fs.W_OK)
+        await fs.ensureDir(this._getRealPath())
```
julien-f: Shouldn't it be `_getFilePath()`?

src/remote-handlers/local.js
```diff
@@ -12,16 +12,20 @@ import {
...
+    return 'file'
+  }
+
+  _getRealPath () {
+    return this._remote.path
   }
 
   _getFilePath (file) {
-    const parts = [this._remote.path]
+    const parts = [this._getRealPath()]
```
julien-f: Use a `const` if you're gonna use it multiple time?

src/remote-handlers/local.js
julien-f: Oh, I think I understand now.

src/remote-handlers/local.js
marsaud: No problem for me, "local" is a user UI feeling name, in the code we do talk about xoa files.

src/remote-handlers/local.js
marsaud: The 2 are clear to me, I meant, "the path on this very local concrete file system" :p

src/remote-handlers/local.js
marsaud: What do you mean ?...

julien-f: You call twice this method in the same scope, you can use a constant value instead.

src/remote-handlers/local.js
marsaud: Ah ok, I did not catch the meaning of your question !

marsaud: Yes that's it, getRealPath is a polymorphic (nfs extends local) method to locate the remote point (mounted or not) on the file system, getFilePath gives the "true" path to a file supposed to be in the remote point.

julien-f:
I'll merge this afet xo-remote-parser has been published.

src/remote-handlers/nfs.js
```diff
@@ -47,7 +47,7 @@ export default class NfsHandler extends LocalHandler {
 
   async _mount () {
     await fs.ensureDir(this._getRealPath())
-    return execa('mount', ['-t', 'nfs', '-o', 'vers=3', `${this._remote.host}:/${this._remote.path}`, this._getRealPath()])
+    return execa('mount', ['-t', 'nfs', '-o', 'vers=3', `${this._remote.host}:${this._remote.path}`, this._getRealPath()])
```
julien-f: It looks better like that :)

Danp2:
@julien-f Shouldn't the NFS handler be using `this._getRealPath()` instead of `remote.path` during the attempt to unmount an NSF share?

julien-f:
You're right, thanks!

PR #353, opened 2016-06-28
A user cannot self-modify permissions

marsaud:


src/api/user.js
```diff
@@ -22,7 +22,7 @@ create.params = {
 // Deletes an existing user.
 async function delete_ ({id}) {
   if (id === this.session.get('user_id')) {
-    throw new InvalidParameters('an user cannot delete itself')
+    throw new InvalidParameters('a user cannot delete itself')
```
julien-f: Are you sure about this?

It feels better but it goes against standard English rules…

src/api/user.js
```diff
@@ -58,6 +58,9 @@ getAll.permission = 'admin'
 // -------------------------------------------------------------------
 
 export async function set ({id, email, password, permission}) {
+  if (permission && id === this.session.get('user_id')) {
+    throw new InvalidParameters('a user cannot modifiy it\'s own permissions')
```
julien-f: `modify` but `change` may be better.

julien-f: `permission`, singular.

src/api/user.js
marsaud: https://www.quora.com/Which-is-the-correct-grammar-usage-a-unique-or-an-unique

julien-f: It feels natural for me too but it's interesting to see the explanation :)

PR #354, opened 2016-06-29
Remote handlers provide a testing method, exposed by xo-server API

marsaud:
Not for merging, yet. For opinion & advice, so far.


src/api/remote.js
```diff
@@ -1,12 +1,12 @@
 export async function getAll () {
-  return /* await */ this.getAllRemotes()
+  return this.getAllRemotes()
```
julien-f: I left the comments for documentation purpose but I guess we can remove them :)

marsaud: Oh, sorry, I thought it looked like something I left myself when I was a bit confused with async/await as a "novelty for me :p

src/remote-handlers/abstract.js
```diff
@@ -47,6 +48,27 @@ export default class RemoteHandlerAbstract {
     throw new Error('Not implemented')
   }
 
+  async test () {
+    const testFileName = `${new Date().valueOf()}.test`
+    const data = (await crypto.randomBytes(1024 * 1024 * 1024)).toString('hex')
```
julien-f: Do not use true random bytes: it's unnecessary and can have a perf impact.

Feel free to `utils/generateUnsecureToken()` into a new `utils/getPseudoRandomBytes()`.

julien-f: Data is much too big (1GiB):

- memory consumption: could be fixed by using a stream and a hash to check the data
- too slow: it can be very long to read and write such a file especially on a slow remote

A 1MiB would be enough.

julien-f: Why converting the data to a hexadecimal string?

src/remote-handlers/abstract.js
```diff
@@ -47,6 +48,27 @@ export default class RemoteHandlerAbstract {
...
+    try {
+      await this.outputFile(testFileName, data)
+      const read = await this.readFile(testFileName)
+      if (read !== data) {
+        throw new Error('Output and input did not match')
```
julien-f: I don't think it's pertinent for a test method to throw an error.

julien-f: My bad, I did not see the surrounding try/catch.

julien-f: Always use lower case messages.

src/remote-handlers/abstract.js
```diff
@@ -47,6 +48,27 @@ export default class RemoteHandlerAbstract {
...
+      }
+      return {
+        result: true,
```
julien-f: `success` instead of `result`.

src/xo-mixins/remotes.js
```diff
@@ -54,6 +54,12 @@ export default class {
     return new Handler[type](remote)
   }
 
+  async testRemote (remote) {
+    const { id = remote } = remote
+    const handler = await this.getRemoteHandler(await this.getRemote(id))
```
julien-f: `Xo#getRemoteHandler()` can handle a remote id directly.

src/remote-handlers/abstract.js
marsaud: What, or where is getPseudoRandomBytes ?...

marsaud: In fact, generateUnsecureToken is enough, no ?...

julien-f: Yep, sorry if I wasn't clear :)

julien-f: You can use directly the random bytes without converting them to base 64.

src/remote-handlers/abstract.js
```diff
@@ -47,6 +48,33 @@ export default class RemoteHandlerAbstract {
     throw new Error('Not implemented')
   }
 
+  async test () {
+    const testFileName = `${new Date().valueOf()}.test`
+    const data = getPseudoRandomBytes(1024 * 1024).toString()
```
julien-f: I don't think it's needed to convert to string, you can read/write buffers directly and compare them using `Buffer#compare()`.

src/remote-handlers/abstract.js
```diff
@@ -47,6 +48,33 @@ export default class RemoteHandlerAbstract {
...
+    let step = 'write'
+    try {
+      await this.outputFile(testFileName, data)
+      step = 'read'
+      const read = await this.readFile(testFileName)
+      if (read !== data) {
+        throw new Error('output and input did not match')
+      }
+      step = 'delete'
+      await this.unlink(testFileName)
```
julien-f: Move this line a `finally` clause or use the `@deferrable` decorator.

src/utils.js
```diff
@@ -182,7 +182,11 @@ export const generateUnsecureToken = (n = 32) => {
     bytes.writeUInt8(Math.random() * 256 | 0, n - 1)
   }
 
-  return base64url(bytes)
+  return bytes
+}
+
+export const generateUnsecureToken = (n = 32) => {
+  return base64url(getPseudoRandomBytes(n))
```
julien-f: Remove the braces and the return :)

src/remote-handlers/abstract.js
```diff
@@ -72,6 +70,12 @@ export default class RemoteHandlerAbstract {
         file: testFileName,
         comment: error.message || String(error)
       }
+    } finally {
+      try {
+        await this.unlink(testFileName)
+      } catch (_) {
+        noop()
+      }
```
julien-f: IMHO it's not necessary to await for the file to be removed: `this.unlink(testFileName).catch(noop)`.

src/remote-handlers/abstract.js
```diff
@@ -47,6 +48,33 @@ export default class RemoteHandlerAbstract {
     throw new Error('Not implemented')
   }
 
+  async test () {
+    const testFileName = `${new Date().valueOf()}.test`
```
julien-f: `Date.now()`

src/remote-handlers/abstract.js
```diff
@@ -47,6 +48,33 @@ export default class RemoteHandlerAbstract {
...
+    try {
+      await this.outputFile(testFileName, data)
+      step = 'read'
+      const read = await this.readFile(testFileName)
+      if (data.compare(read) !== 0) {
+        throw new Error('output and input did not match')
+      }
+      return {
+        success: true,
+        comment: 'Test Successful'
```
julien-f: No comments needed.

src/remote-handlers/abstract.js
```diff
@@ -47,6 +48,33 @@ export default class RemoteHandlerAbstract {
...
+      }
+    } catch (error) {
+      return {
+        success: false,
+        step,
+        file: testFileName,
+        comment: error.message || String(error)
```
julien-f: `comment` → `error`.

PR #356, opened 2016-07-07
Fix patch parsing in XML

marsaud:
vatesfr/xo-web#1236


marsaud:
Do not merge, the bug is elsewhere

PR #363, opened 2016-07-19
feat(schedules): schedules support timezones (close vatesfr/xo-web#1258)

Wescoeur:


package.json
```diff
@@ -79,6 +79,8 @@
     "make-error": "^1",
     "micromatch": "^2.3.2",
     "minimist": "^1.2.0",
+    "moment": "^2.14.1",
```
julien-f: Unnecessary.

julien-f:
Add a comment explaining why it's correct to interpret cron patterns and display date in UTC.

julien-f:
My bad, it's for xo-web's related PR.

src/api.js
```diff
@@ -188,6 +189,14 @@ methodSignature.description = 'returns the signature of an API method'
 
 // ===================================================================
 
+export function getTimezone () {
+  return moment.tz.guess()
+}
```
julien-f: Do it only once at startup:

```
const getServerTimezone = (tz => () => tz)(moment.tz.guess())
```

julien-f: Do not export this function.

julien-f: Name this function `getServerTimezone`.

src/utils.js
```diff
@@ -12,6 +12,10 @@ import keys from 'lodash/keys'
 import kindOf from 'kindof'
 import multiKeyHashInt from 'multikey-hash'
 import xml2js from 'xml2js'
+/* Moment timezone can be loaded only one time, it's a workaround to load
+   the latest version because cron module uses an old version of moment which
+   does not implement `guess` function for example. */
+import moment from 'moment-timezone' // eslint-disable-line no-unused-vars
```
julien-f: If you don't need to use the value, remove its identifier from the import clause:

```js
import 'moment-timezone'
```

julien-f: Style:

- Add blank lines around this import block
- use single line comment syntax

PR #366, opened 2016-07-24
add dockerfile and docker compose

hkparker:
Hi XO team, love your work.

Not sure this is the best place for these but I wanted to share my recently docker success.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM node:6-wheezy@sha256:5eb189fa5041cb692146d3212593ef965e8adf58f48a2bfc738fa30bb7f5045f
+
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install build-essential libpng-dev git python-minimal -y
+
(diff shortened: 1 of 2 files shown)


olivierlambert:
Hi @hkparker,

Thanks for your contribution!

We are not Docker expert, thus I got a question: why not creating a Docker stuff right in DockerHub (eg like here: https://hub.docker.com/r/sammcj/docker-xen-orchestra/)? I mean, if it's possible to avoid adding extra file here, that would be better. But as I said, we are not Docker expert so maybe we missed the point.

hkparker:
It's not a bad idea to keep the docker files in a separate repository, and to be honest I didn't know there was already a community container.  I'm totally happy to make my own repo and push to dockerhub, if you don't mind.  The only reason you'd have for including these files in a repository of your own is if you wanted to maintain an official container.  Happy to help if you are interested in doing that ever, thanks for working on xo!

olivierlambert:
@hkparker thanks for the clarification :) We are happy about the community creating stuff around XO, but we can't maintain these ourselves (because we are focusing on providing support for XOA)

So feel free to create a community container as you like!

PR #368, opened 2016-07-25
feat(models/schedule): cast `null` to null (vatesfr/xo-web#1314)

Wescoeur:


src/models/schedule.js
```diff
@@ -34,6 +34,8 @@ export class Schedules extends Collection {
     const schedules = await super.get(properties)
     forEach(schedules, schedule => {
       schedule.enabled = (schedule.enabled === 'true')
+      const { timezone } = schedule
+      schedule.timezone = (timezone === 'null' ? null : timezone)
```
julien-f: Simply remove the timezone field if there is no timezone.

src/xo-mixins/scheduling.js
```diff
@@ -155,7 +155,9 @@ export default class {
     if (cron) schedule.set('cron', cron)
     if (enabled !== undefined) schedule.set('enabled', enabled)
     if (name !== undefined) schedule.set('name', name)
-    if (timezone !== undefined) schedule.set('timezone', timezone)
+    if (timezone !== undefined) {
+      schedule.set('timezone', timezone || undefined)
```
julien-f:
```js
if (timezone === null) {
  schedule.set('timezone', undefined) // Remove current timezone
} else if (timezone !== undefined) {
  schedule.set('timezone', timezone)
}
```

PR #369, opened 2016-07-27
feat(job-executor): supports dynamic parameters (vatesfr/xo-web#837)

Wescoeur:


src/job-executor.js
```diff
@@ -18,29 +17,91 @@ export class UnsupportedVectorType extends JobExecutorError {
...
+  for (let i = 0; i < m; ++i) {
+    const value = vector[i]
+
+    _combine(vectors, nLast, (vector) => {
+      vector.push(value)
+      cb(vector)
+    })
+  }
 }
+const combine = vectors => cb => _combine(vectors, vectors.length, cb)
```
julien-f: Would be nice to add a small explanation.

src/job-executor.js
```diff
@@ -18,29 +17,91 @@ export class UnsupportedVectorType extends JobExecutorError {
...
+
+export const vectorToObject = vector => {
```
julien-f: Small explanation.

src/job-executor.js
```diff
@@ -18,29 +17,91 @@ export class UnsupportedVectorType extends JobExecutorError {
...
+  const obj = {}
+  const n = vector.length
+  for (let i = 0; i < n; ++i) {
+    assign(obj, vector[i])
+  }
+
+  return obj
+}
+
+export const crossProduct = vectors => cb => combine(vectors)(vector => {
```
julien-f: Small explanation.

src/job-executor.js
```diff
@@ -18,29 +17,91 @@ export class UnsupportedVectorType extends JobExecutorError {
...
+  cb(vectorToObject(vector))
+})
+
+// ===================================================================
+
+export const thunkToArray = thunk => {
```
julien-f: Small explanation.

julien-f: All the methods above are completely generic and should be in a dedicated module.

src/job-executor.js
```diff
@@ -18,29 +17,91 @@ export class UnsupportedVectorType extends JobExecutorError {
...
+  const values = []
+  thunk(::values.push)
+  return values
+}
+
+// ===================================================================
+
+const getNodeValues = node => node.values || node.items
```
julien-f: Nope, it should be deterministic.

src/job-executor.js
julien-f: Not sure this function is really necessary as it can expressed very simply: `assign({}, ...objects)`.

If you wan to keep it, I think it would be better named `mergeObjects()`.

src/job-executor.js
julien-f: If you want to make it generic, add a second param `mergeFn = mergeObjects`.

src/api/job.js
```diff
@@ -27,6 +30,7 @@ create.params = {
   job: {
     type: 'object',
     properties: {
+      id: {type: 'string'},
```
julien-f: ?

src/api/job.js
```diff
@@ -39,11 +43,20 @@ create.params = {
             type: 'array',
             items: {
               type: 'object',
-              properties: {
-                type: {type: 'string'},
-                values: {
-                  type: 'array',
-                  items: {type: 'object'}
+              exec: function (schema, post) {
```
julien-f: I prefer keeping JSON schemas, avoid using functions.

src/job-executor.js
```diff
@@ -18,30 +23,84 @@ export class UnsupportedVectorType extends JobExecutorError {
...
+  },
+  fetchObjects (node) {
+    const { pattern } = node
+
+    const predicate = object => {
+      for (const key in pattern) {
+        const value = pattern[key]
+        const objectValue = object[key]
+
+        if (!Array.isArray(value)) {
```
julien-f: `lodash/isArray`.

src/api/job.js
Wescoeur: Sorry. Copy-paste is a bad practice. (`set` method => `create` method) :sob:

src/job-executor.js
```diff
@@ -18,30 +27,60 @@ export class UnsupportedVectorType extends JobExecutorError {
...
+  extractProperties ({ pattern, value }) {
+    return mapValues(pattern, key => value[key])
+  },
+  crossProduct ({ items }) {
+    return thunkToArray(crossProduct(
+      map(items, value => resolveParamsVector.call(this, value))
+    ))
+  },
+  fetchObjects (node) {
+    const { pattern } = node
```
julien-f: Move destructuring in param list.

src/job-executor.js
```diff
@@ -18,30 +27,60 @@ export class UnsupportedVectorType extends JobExecutorError {
...
+    return filter(this.xo.getObjects(), object => match(pattern, object))
+  },
+  map ({ collection, iteratee, iterateeArgs }) {
+    return map(resolveParamsVector.call(this, collection), value =>
+      resolveParamsVector.call(this, { type: iteratee, ...iterateeArgs, value })
+    )
+  },
+  set: ({ values }) => values
```
julien-f: Are all this methods used?

src/math.js
```diff
@@ -0,0 +1,43 @@
...
+  for (let i = 0; i < m; ++i) {
+    const value = vector[i]
+
+    _combine(vectors, nLast, (vector) => {
+      vector.push(value)
+      cb(vector)
+    })
+  }
+}
+const combine = vectors => cb => _combine(vectors, vectors.length, cb)
```
julien-f: Little explaination like the other functions please :)

src/utils.js
```diff
@@ -478,3 +478,12 @@ export const scheduleFn = (cronTime, fn, timeZone) => {
 
 // Wrap a value in a function.
 export const wrap = value => () => value
+
+// -------------------------------------------------------------------
+
+// Create an array which contains the results of one thunk function.
+export const thunkToArray = thunk => {
```
julien-f: Keep the exports sorted :)

src/job-executor.js
Wescoeur: Yes.

src/utils.js
julien-f: Add a small comment explaning this function only works with synchronous thunks.

src/job-executor.js
```diff
@@ -18,30 +27,59 @@ export class UnsupportedVectorType extends JobExecutorError {
...
+  },
+  crossProduct ({ items }) {
+    return thunkToArray(crossProduct(
+      map(items, value => resolveParamsVector.call(this, value))
+    ))
+  },
+  fetchObjects ({ pattern }) {
+    return filter(this.xo.getObjects(), object => match(pattern, object))
+  },
+  map ({ collection, iteratee, iterateeArgs }) {
```
julien-f: Merge `iteratee` and `iterateeArgs`:

```js
{
  type: 'map',
  iteratee: {
    type: 'extractProperties',
    pattern: { id: 'id' }
  }
}
```

julien-f: Add a comment to say that in the future, an additional parameter may be added to configure the name of the injected parameter.

src/job-executor.js
```diff
@@ -18,30 +27,59 @@ export class UnsupportedVectorType extends JobExecutorError {
...
+    return isArray(value) && every(pattern, subpattern =>
+      some(value, subvalue => match(subpattern, subvalue))
+    )
+  }
+
+  return pattern === value
+}
+
+const paramsVectorActionsMap = {
+  extractProperties ({ pattern, value }) {
```
julien-f: `pattern` → `mapping`.

src/job-executor.js
```diff
@@ -18,30 +27,67 @@ export class UnsupportedVectorType extends JobExecutorError {
...
+      map(items, value => resolveParamsVector.call(this, value))
+    ))
+  },
+  fetchObjects ({ pattern }) {
+    return filter(this.xo.getObjects(), object => match(pattern, object))
+  },
+  map ({ collection, iteratee }) {
+    return map(resolveParamsVector.call(this, collection), value => {
+      const {
+        paramName = 'value',
```
julien-f: `paramName` is not in `iteratee` but directly in the `node`!

PR #385, opened 2016-09-01
feat(network): add defaultLockingMode to API

olivierlambert:
Related to https://github.com/vatesfr/xo-web/issues/1427


src/xapi-object-to-xo.js
```diff
@@ -483,6 +483,7 @@ const TRANSFORMS = {
   network (obj) {
     return {
       bridge: obj.bridge,
+      defaultLockingMode: obj.default_locking_mode === 'disabled',
```
julien-f: Seems weird for a *mode*, a string would be expected, not a boolean: either change the property name or the value.

src/api/network.js
```diff
@@ -31,11 +31,13 @@ export async function set ({
 
   name_description: nameDescription,
   name_label: nameLabel,
+  defaultLocking,
```
julien-f: Rename.

PR #393, opened 2016-09-14
remove node-smb2 dependency

nraynaud:
it is indirectly used by @marsaud/smb2-promise


julien-f:
It was added to bump the version of node-smb2 without needing to publish a new @marsaud/smb2-promise, it is no longer needed now.

Thanks :)

PR #401, opened 2016-09-24
Create filelock while backup is being written

bcdonadio:
This PR creates a simple filelock on the remote just before the backup output file is created and delete it when the operation finishes (both on error and success). Additionally, it deletes the partial output file if the export is not successful.

Rationale: On my use case, I have a staging NFS mount where XO saves its backups. When the save is completed, I need to compress it (I don't want to run the compressor on the same system as XO) and then send it to S3. However, on XO current implementation I have no way to know if the backup file is still being written, potentially leading to corrupted backups.

Suggestions on improvements or alternative ways to achieve this are welcome.


fufroma:
If XO is abruptly restarted, should the` .lock` file be cleaned up with the interrupted backup file?

bcdonadio:
I believe that this behaviour can result in problems if you have more than one XO instance running on the same Xen pool (but I'm not sure if this multiplicity is supported in the first place). OTOH, maybe this is a good anti-entropy practice...

Maybe should we clean the stale lock files and interrupted backup files only when running the next job matching the same tag?

julien-f:
Why not simply change the initial name of the file (e.g. prefixing with a dot `.`) and rename it on success?

bcdonadio:
Holy crap. Indeed much simpler, lol. I will rewrite with rename as soon as I can.

But the question remains, should we delete interrupted backup files automatically?

julien-f:
Yes, we should probably delete them because they are unusable and because otherwise they might never get removed and clutter the file system.

fufroma:
I'm not a fan of the `.` prefix. It's an hidden file in Linux, there is no interest in hidding this file here. IMHO we _want_ to see there are backup in progress.

Adding a .temp suffix could do the same trick, but without hidding it.

Danp2:
Would this be better implemented as an XO plugin?

fufroma:
> Would this be better implemented as an XO plugin?

I really think this should be mainstream.

julien-f:
@Danp2 I aggree with @fufroma that this should be done directly in XO, and it's far easier to do :)

Do you see any issue with this?

Danp2:
@julien-f No, I don't see an issue with the proposed solution. My idea was to create a more generic solution that could kick off a post-job task.

julien-f:
@Danp2 I'm open to suggestions, but it should be fairly simple and portable.

julien-f:
@bcdonadio Are you still working on this? :)

julien-f:
Closing this, new backup code will use temporary filenames during the backup process.

PR #406, opened 2016-09-29
Chain the backup delta VHD files

nraynaud:
The goal is for a tool like vhdimount to be able to mount any file and use it as a disk to recover specific file in it.
I forked struct-fu to add support for utf16be, but the original author is not thrilled: https://github.com/natevw/struct-fu/pull/6 There might be a decision to be made there before merging.


package.json
```diff
@@ -93,7 +93,7 @@
     "semver": "^5.1.0",
     "serve-static": "^1.9.2",
     "stack-chain": "^1.3.3",
-    "struct-fu": "^1.0.0",
+    "@nraynaud/struct-fu": "^1.0.1",
```
julien-f: Would you mind also adding me as a GitHub/npm maintainer (increase the bus factor :) ).

src/xo-mixins/backups.js
```diff
@@ -291,6 +291,14 @@ export default class {
     return backups.slice(i)
   }
 
+  // fix the parent UUID and filename in delta files after download from xapi or backup compression
+  async _chainDeltaVdiBackups ({handler, dir}) {
```
julien-f: Could not this be directly integrated with the merge process?

src/xo-mixins/backups.js
```diff
@@ -291,6 +291,14 @@ export default class {
...
+    const backups = await this._listVdiBackups(handler, dir)
+    for (let i = 1; i < backups.length; i++) {
+      await chainVhd(handler, dir + '/' + backups[i - 1], handler, dir + '/' + backups[i])
+    }
```
julien-f: Is it necessary to do this for all entries?
Is it not done for each delta everytime it's created?

(Real questions, I may be completely wrong.)

nraynaud: There are 2 cases that need adjusting: the top of the delta stack when a new backup as been downloaded, and the first delta after a merge.
There is also the modification of the backup stacks saved before this change was made.

src/xo-mixins/backups.js
nraynaud: A new backup delta will probably not be merged, and after a merge we need to change the first non-merged delta. So we are impacting delta that won't be touched by the merge.

package.json
nraynaud: I will, but wouldn't you prefer forking it in the Vates organization?

olivierlambert: That would be better

olivierlambert: @nraynaud https://github.com/vatesfr/struct-fu

src/xo-mixins/backups.js
julien-f: Ok, I agree it's far simpler to iterate over all deltas.

PR #407, opened 2016-09-29
feat(network): create a bonded network

pdonias:
See vatesfr/xo-web#876


src/xapi-object-to-xo.js
```diff
@@ -406,6 +406,7 @@ const TRANSFORMS = {
       type: 'PIF',
 
       attached: Boolean(obj.currently_attached),
+      bondMasterOf: obj.bond_master_of,
```
julien-f: Bond objects are not available in the API, so it has very little value exposing this id.

Make it a boolean and rename it.

src/xapi/index.js
```diff
@@ -1868,11 +1869,29 @@ export default class Xapi extends XapiBase {
...
 
+  async createBondedNetwork ({
+    name,
+    description,
+    pifIds,
+    mtu,
+    mac,
+    bond_mode
+  }) {
+    const network = await this.createNetwork({ name, description, mtu })
```
julien-f: Use object splat and spread to forward all unused params to `createNetwork()`.

src/xapi/index.js
```diff
@@ -1868,11 +1869,29 @@ export default class Xapi extends XapiBase {
...
+    await this.call('Bond.create', network.$ref, map(pifIds, pifId => this.getObject(pifId).$ref), mac, bond_mode)
+
+    return network
+  }
+
   async deleteNetwork (networkId) {
     const network = this.getObject(networkId)
     await Promise.all(
-      mapToArray(network.$PIFs, (pif) => this.call('VLAN.destroy', pif.$VLAN_master_of.$ref))
+      mapToArray(network.$PIFs, pif => pif.$VLAN_master_of && this.call('VLAN.destroy', pif.$VLAN_master_of.$ref))
```
julien-f: Use an intermediary variable for `$VLAN_master_of`.

src/xapi/index.js
```diff
@@ -1868,11 +1869,29 @@ export default class Xapi extends XapiBase {
...
     )
+    const bond = !isEmpty(network.$PIFs) && network.$PIFs[0].bond_master_of[0]
```
julien-f: Intermediary variable for pifs.

src/xapi/index.js
```diff
@@ -1868,11 +1869,29 @@ export default class Xapi extends XapiBase {
...
+    if (bond) {
+      await this.call('Bond.destroy', bond)
```
julien-f: You should probably use `map()` here.

src/xapi/index.js
```diff
@@ -1868,11 +1869,29 @@ export default class Xapi extends XapiBase {
...
+  async createBondedNetwork ({
+    name,
+    description,
+    pifIds,
+    mtu,
+    mac,
+    bond_mode
+  }) {
+    const network = await this.createNetwork({ name, description, mtu })
+    await this.call('Bond.create', network.$ref, map(pifIds, pifId => this.getObject(pifId).$ref), mac, bond_mode)
```
julien-f: We are creating a bond for a single host and XAPI should create all the other bonds automatically.

Add a comment explaining that and that we are not sure about it.

src/xapi/index.js
julien-f: I fear that we may try to delete the same VLAN multiple times.

julien-f: Better use this simplification rule: `obj.$prop.$ref` →`obj.prop`.

src/xapi/index.js
julien-f: Do we need to also iterate over all PIFs?

src/api/network.js
```diff
@@ -61,7 +54,7 @@ createBonded.params = {
     }
   },
   mtu: { type: ['integer', 'string'], optional: true },
-  bond_mode: { type: 'string' }
+  bondMode: { type: 'string', pattern: new RegExp(`^(${getBondModes().join('|')})$`) }
```
julien-f: Maybe add a comment saying there are no enums in schema-inspector.

src/xapi/index.js
```diff
@@ -1870,28 +1871,33 @@ export default class Xapi extends XapiBase {
...
+    await this.call('Bond.create', network.$ref, map(pifIds, pifId => this.getObject(pifId).$ref), mac, bondMode)
 
     return network
   }
 
   async deleteNetwork (networkId) {
     const network = this.getObject(networkId)
+    const pifs = network.$PIFs
+
+    const vlans = uniq(mapToArray(pifs, pif => pif.VLAN_master_of))
```
julien-f: What does it do when there `VLAN_master_of` is *empty*.

src/xapi/index.js
slindaqiu: Where is the definition of 'Bond.create'?

olivierlambert: @slindaqiu http://xapi-project.github.io/xen-api/classes/bond.html

PR #410, opened 2016-10-03
fix(utils/makeEditObject): better constraints handle

pdonias:


src/xapi/utils.js
```diff
@@ -317,14 +317,21 @@ export const makeEditObject = specs => {
...
         forEach(constraints, (constraint, constraintName) => {
-          // This constraint value is already defined: bypass the constraint.
-          if (values[constraintName] != null) {
-            return
-          }
-
-          if (!constraint(specs[constraintName].get(object), value)) {
-            const cb = set(value, constraintName)
-            cbs.push(cb)
+          /* Before setting a property to a new value, if the constraint check fails (e.g. memoryMin > memoryMax):
```
julien-f: Use single line comments.

src/xapi/utils.js
```diff
@@ -317,14 +317,21 @@ export const makeEditObject = specs => {
...
+           *        constraint <-- constraintNewValue THEN property <-- value (e.g. memoryMax <-- 2048 THEN memoryMin <-- 1024)
+           *    - if the user DOES NOT want to set the constraint (ie constraintNewValue is NOT defined):
+           *        constraint <-- value THEN property <-- value (e.g. memoryMax <-- 1024 THEN memoryMin <-- 1024)
+           */
+          const constraintCurrentValue = specs[constraintName].get(object)
+          const constraintNewValue = values[constraintName]
+          if (!constraint(constraintCurrentValue, value)) {
+            const cb = constraintNewValue == null
+              ? set(value, constraintName)
+              : set(constraintNewValue, constraintName)
```
julien-f: Factorize more.

src/xapi/utils.js
```diff
@@ -317,14 +317,21 @@ export const makeEditObject = specs => {
...
+            if (cb) {
```
julien-f: Why this new test?

src/xapi/utils.js
```diff
@@ -317,14 +317,21 @@ export const makeEditObject = specs => {
...
-
-          if (!constraint(specs[constraintName].get(object), value)) {
-            const cb = set(value, constraintName)
-            cbs.push(cb)
+          /* Before setting a property to a new value, if the constraint check fails (e.g. memoryMin > memoryMax):
+           *    - if the user wants to set the constraint (ie constraintNewValue is defined):
+           *        constraint <-- constraintNewValue THEN property <-- value (e.g. memoryMax <-- 2048 THEN memoryMin <-- 1024)
+           *    - if the user DOES NOT want to set the constraint (ie constraintNewValue is NOT defined):
+           *        constraint <-- value THEN property <-- value (e.g. memoryMax <-- 1024 THEN memoryMin <-- 1024)
+           */
```
julien-f: Explain that, in certain scenarii, there are duplicate requests but, even though it's not perfect, it's good enough for now.

src/xapi/utils.js
pdonias: If `set` returns `undefined` for one of the constraints (e.g.: if the constraint property is being set to the same value as the current one), I don't want `undefined` to be pushed in the callbacks array because it will throw an error when it will be called.

julien-f: Yes, but why this test did not exist before?

PR #424, opened 2016-10-20
feat(pif): expose IP config modes

pdonias:
See #1651


src/api/pif.js
```diff
@@ -1,5 +1,12 @@
 // TODO: too low level, move into host.
 
+export function getIpv4ConfigurationModes () {
+  return ['None', 'DHCP', 'Static']
```
julien-f: `Xapi.IPV4_CONFIG_MODES` & XAPI.IPV6_CONFIG_MODES`.

PR #426, opened 2016-10-21
feat(pif): editVlan

pdonias:
See vatesfr/xo-web#1092


src/api/pif.js
```diff
@@ -76,3 +76,18 @@ reconfigureIp.params = {
 reconfigureIp.resolve = {
   pif: ['id', 'PIF', 'administrate']
 }
+
+// ===================================================================
+
+export async function editVlan ({ pif, vlan }) {
```
julien-f: IMHO, this name is incorrect, it should be named `editPif()`.

src/xapi/index.js
```diff
@@ -1874,6 +1874,24 @@ export default class Xapi extends XapiBase {
     return this._getOrWaitObject(networkRef)
   }
 
+  async editPif ({
+    pifId,
```
julien-f: For homogeneity, `pifId` should be an independent first param.

src/xapi/index.js
```diff
@@ -1874,6 +1874,24 @@ export default class Xapi extends XapiBase {
...
+    vlan
+  }) {
+    const pif = this.getObject(pifId)
+    await this.call('VLAN.destroy', pif.VLAN_master_of)
+    const physPif = find(this.objects.all, obj => (
```
julien-f: Move this before any side effects occur and throw an error if this object is not found.

PR #448, opened 2016-11-14
feat(job#create,job#set): userId parameter

pdonias:
See vatesfr/xo-web#1733

### DIFF
--- a/src/api/job.js
+++ b/src/api/job.js
@@ -18,7 +18,9 @@ get.params = {
 }
 
 export async function create ({job}) {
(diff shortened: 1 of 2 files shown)


src/api/job.js
```diff
@@ -51,7 +54,11 @@ create.params = {
 }
 
 export async function set ({job}) {
-  await this.updateJob(job)
+  if (!job.userId) {
+    job.userId = this.session.get('user_id')
+  }
```
julien-f: It should be possible to not change a job's user.

PR #449, opened 2016-11-15
feat(api): map 10 XAPI errors to `XoError`s

pdonias:
See vatesfr/xo-web#1481


src/xo-mixins/api.js
```diff
@@ -303,7 +304,7 @@ export default class Api {
 
       const xoError = XAPI_ERROR_TO_XO_ERROR[error.code]
       if (xoError) {
-        throw xoError(error.params)
+        throw xoError(error.params, ref => this._xo.getObject(ref).id)
```
julien-f: What if the object could not be found?

PR #451, opened 2016-11-16
Undefinded data

badrAZ:


src/xo-mixins/plugins.js
```diff
@@ -251,6 +251,6 @@ export default class {
       }
     }
 
-    await plugin.instance.test(data)
+    await plugin.instance.test(data || undefined)
```
julien-f: What does this change mean?

badrAZ: It worked without || undefined

PR #455, opened 2016-11-22
update noVNC to latest upstream version

nraynaud:
don't forget the matching xo-web PR:
https://github.com/vatesfr/xo-web/pull/1780


julien-f:
@nraynaud see https://github.com/vatesfr/xo-web/pull/1780#issuecomment-288342186

Should be pretty easy on this side, I can do it if you want.

PR #457, opened 2016-11-22
fix(vif.set): remove old VIF before creating new one

pdonias:


src/api/vif.js
```diff
@@ -85,16 +84,16 @@ export async function set ({
     attached == null && (attached = vif.attached)
 
     await this.allocIpAddresses(vif.id, null, oldIpAddresses)
+    await xapi.deleteVif(vif._xapiId)::pCatch(noop)
```
julien-f: Do not ignore errors.

julien-f: Try to delete the VIF before deallocating its IPs.

PR #459, opened 2016-12-01
feat(job-executor.match): __not pattern property

pdonias:
See vatesfr/xo-web#1503


src/job-executor.js
```diff
@@ -32,8 +32,13 @@ export class UnsupportedVectorType extends JobExecutorError {
...
 const match = (pattern, value) => {
   if (isPlainObject(pattern)) {
-    if (pattern.__or && size(pattern) === 1) {
-      return some(pattern.__or, subpattern => match(subpattern, value))
+    if (size(pattern) === 1) {
+      if (pattern.__or) {
+        return some(pattern.__or, subpattern => match(subpattern, value))
+      }
+      if (pattern.__not) {
+        return !some(pattern.__not, subpattern => match(subpattern, value))
```
julien-f: Keep the semantic simple: `return !match(pattern.__not, value)`.

PR #472, opened 2017-01-05
feat(server): store connection error in database

pdonias:
See vatesfr/xo-web#1833


src/xo-mixins/xen-servers.js
```diff
@@ -278,7 +282,10 @@ export default class {
 
     xapi.xo.install()
 
-    await xapi.connect()
+    await xapi.connect().then(
+      () => this.updateXenServer(id, { error: '' }),
+      error => this.updateXenServer(id, { error: JSON.stringify(error) })
```
julien-f: Use `utils/serializeError`.

PR #479, opened 2017-01-10
feat(patching): eject tools ISOs before patching host

pdonias:
Fixes #1798


src/xapi/mixins/patching.js
```diff
@@ -149,6 +151,30 @@ export default {
     )
   },
 
+  async _ejectToolsIsos (hostRef) {
+    const ejections = []
+
+    forEach(
```
julien-f: Use `map`.

julien-f: If you want to optimize, you can use `mapFilter`.

Move the implementation from `xapi/utils` to `utils`.

src/xapi/mixins/patching.js
```diff
@@ -152,27 +153,22 @@ export default {
...
-        (!hostRef || object.resident_on === hostRef)
-      ),
+    return Promise.all(mapFilter(
+      this.objects.all,
       vm => {
+        if (vm.$type !== 'vm' || (hostRef && vm.resident_on !== hostRef)) {
+          return
+        }
+
         const shouldEjectCd = some(vm.VBDs, vbdRef => {
```
julien-f: `vm.$VBDs`.

src/xapi/mixins/patching.js
```diff
@@ -152,27 +153,22 @@ export default {
...
           const vdiRef = this.getObject(vbdRef).VDI
```
julien-f: `const vdi = vbd.$VDI`.

PR #483, opened 2017-01-13
fix: storing directory

badrAZ:


src/index.js
```diff
@@ -223,7 +224,12 @@ async function registerPlugin (pluginPath, pluginName) {
   // The default export can be either a factory or directly a plugin
   // instance.
   const instance = isFunction(factory)
-    ? factory({ xo: this })
+    ? factory({
+      xo: this,
+      dir: () => {
+        const dir = `${this._config.datadir}/${pluginName}`
+        return ensureDir(dir).then(() => ({dir}))
```
julien-f: No need to return an object.

src/index.js
```diff
@@ -223,7 +224,12 @@ async function registerPlugin (pluginPath, pluginName) {
   // The default export can be either a factory or directly a plugin
   // instance.
   const instance = isFunction(factory)
-    ? factory({ xo: this })
+    ? factory({
+      xo: this,
+      dir: () => {
```
julien-f: The name is not meaningful enough.

src/index.js
```diff
@@ -226,9 +226,9 @@ async function registerPlugin (pluginPath, pluginName) {
   const instance = isFunction(factory)
     ? factory({
       xo: this,
-      dir: () => {
+      getDataDir: () => {
         const dir = `${this._config.datadir}/${pluginName}`
-        return ensureDir(dir).then(() => ({dir}))
+        return ensureDir(dir).then(() => (dir))
```
julien-f: You can remove the parens.

PR #487, opened 2017-01-19
feat(host.installSupplementalPack)

pdonias:
See vatesfr/xo-web#1460


src/api/host.coffee
```diff
@@ -260,6 +262,51 @@ stats.resolve = {
...
+  return {
+    $sendTo: yield @registerHttpRequest(handleInstallSupplementalPack, { host, size })
+  }
+
+installSupplementalPack.description = 'installs supplemental pack from ISO file'
+
+installSupplementalPack.params = {
+  host: { type: 'string' }
+  size: { type: 'number' }
+}
```
julien-f: Not necessary to supply the size: you can get it in the handler from the `content-length` header.

src/api/host.coffee
```diff
@@ -260,6 +262,51 @@ stats.resolve = {
...
+  if not sr
+    pbds = @getObject(hostId, 'host').$PBDs
+
+    sr = $find(
+      mapToArray(pbds , (pbd) => @getObject(@getObject(pbd, 'PBD').SR))
+      (sr) -> sr && sr.content_type == 'user' && sr.size - sr.physical_usage >= size
+    )
+
+  if not sr
+    throw new Error('no SR available to store installation file')
```
julien-f: Why not handling the SR directly in `Xapi#installSupplementalPack`?

src/xapi/index.js
```diff
@@ -1105,6 +1105,22 @@ export default class Xapi extends XapiBase {
     return loop()
   }
 
+  async installSupplementalPack (stream, { srId, hostId, size, xapi }) {
+    const sr = this.getObject(srId)
+    const host = this.getObject(hostId)
+
+    const vdi = await xapi.createVdi(size, {
+      sr: sr.$ref,
+      name_label: 'Supplemental pack ISO',
```
julien-f: Add `[XO]` prefix.

src/api/host.coffee
pdonias: To keep `Xapi#installSupplementalPack` as simple as possible and to pass as few arguments as possible through `registerHttpRequest`.
Is it better to handle it in `Xapi#installSupplementalPack` anyway?

julien-f: Yes, because we could use it without having to duplicate this logic.

src/xapi/index.js
```diff
@@ -1105,18 +1105,29 @@ export default class Xapi extends XapiBase {
     return loop()
   }
 
-  async installSupplementalPack (stream, { srId, hostId, size, xapi }) {
-    const sr = this.getObject(srId)
-    const host = this.getObject(hostId)
+  async installSupplementalPack (stream, { hostId }) {
+    const fileSize = stream.get('content-length')
```
julien-f: Use `stream.length`.

src/api/host.coffee
```diff
@@ -264,45 +264,31 @@ exports.stats = stats;
...
-  if not sr
-    throw new Error('no SR available to store installation file')
-
   # Timeout seems to be broken in Node 4.
   # See https://github.com/nodejs/node/issues/3319
   req.setTimeout(43200000) # 12 hours
 
   try
-    yield xapi.installSupplementalPack(req, { srId: sr.$id, hostId, size, xapi })
+    yield xapi.installSupplementalPack(req, { hostId })
```
julien-f: Add a `length` property on `req` from `req.headers['content-length']`.

julien-f: Handle the case were this header is missing.

src/api/host.coffee
```diff
@@ -260,6 +262,38 @@ stats.resolve = {
...
+installSupplementalPack = $coroutine ({host}) ->
+  return {
+    $sendTo: yield @registerHttpRequest(handleInstallSupplementalPack, { host })
+  }
+
+installSupplementalPack.description = 'installs supplemental pack from ISO file'
+
+installSupplementalPack.params = {
+  host: { type: 'string' }
+}
```
julien-f: Add a resolve to check for the `admin` permission.

src/api/host.coffee
```diff
@@ -260,6 +262,42 @@ stats.resolve = {
...
+    res.end(format.response(0))
+  catch e
+    res.writeHead(500)
+    res.end(format.error(0, new Error(e.message)))
+
+  return
+
+installSupplementalPack = $coroutine ({host}) ->
+  return {
+    $sendTo: yield @registerHttpRequest(handleInstallSupplementalPack, { host })
```
julien-f: Pass the host id instead.

PR #505, opened 2017-02-23
feat(XOSAN)

pdonias:


src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
+import defer from 'golike-defer'
+import fromPairs from 'lodash/fromPairs'
+import fs from 'fs-promise'
+import map from 'lodash/map'
+import arp from 'arp-a'
+import { spawn } from 'child_process'
+import {
```
julien-f: Add a blank line above to separate global and local imports.

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
+import defer from 'golike-defer'
+import fromPairs from 'lodash/fromPairs'
+import fs from 'fs-promise'
+import map from 'lodash/map'
+import arp from 'arp-a'
```
julien-f: Sort your imports alphabetically.

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+      stderr += buffer.toString()
+    })
+    child.on('close', (code) => {
+      if (code !== 0) {
+        reject(stderr)
+      } else {
+        resolve(resp)
+      }
+    })
+  })
```
julien-f: Use `execa` instead :)

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+   cluster.eager-lock: enable
+   performance.io-cache: off
+   performance.read-ahead: off
+   performance.quick-read: off
+   performance.stat-prefetch: on
+   performance.strict-write-ordering: off
+   cluster.server-quorum-type: server
+   cluster.quorum-type: auto
+   */
+  let info = fromPairs(result.trim().split('\n')
```
julien-f: Use `split-lines`.

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+    .map(line => line.split(':').map(val => val.trim()))
+    // some lines have more than one ':', re-assemble them
+    .map(line => [line[0], line.slice(1).join(':')]))
```
julien-f: You can use the following function (put into `utils`):

```js
const splitFirst = (string, separator) => {
  const i = string.indexOf(separator)
  return i === -1 ? null : [
    string.slice(0, i),
    string.slice(i + separator.length)
  ]
}
```

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+  let getNumber = item => parseInt(item.substr(5))
```
julien-f: Use prefix operator `+` to cast to number, it's more robust than `parseInt()`

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+  let brickKeys = Object.keys(info).filter(key => key.match(/^Brick[1-9]/)).sort((i1, i2) => getNumber(i1) - getNumber(i2))
+  // expected brickKeys : [ 'Brick1', 'Brick2', 'Brick3' ]
+  info['Bricks'] = brickKeys.map(key => {
+    const ip = info[key].split(':')[0]
+    return { config: info[key], ip: ip, vm: giantIPtoVMDict[ip] }
+  })
+  await new Promise((resolve, reject) => arp.table((err, entry) => {
```
julien-f: Why do you need a promise? It does not look async :/

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+
+getVolumeInfo.params = {
+  sr: {
+    type: 'string'
+  }
+}
+getVolumeInfo.resolve = {
+  sr: ['sr', 'SR', 'administrate']
+}
+function trucate2048 (value) {
```
julien-f: IMHO, `floor2048()` would be more correct.

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+  let vm = vmAndParam.vm
+  // refresh the object so that sizes are correct
+  const params = vmAndParam.params
+  const ip = params.xenstore_data['vm-data/ip']
+  const sr = xapi.getObject(params.sr.$id)
+  await xapi._waitObjectState(sr.$id, sr => Boolean(sr.$PBDs))
+  let host = xapi.getObject(xapi.getObject(sr.$PBDs[0]).host)
+  let firstVif = vm.$VIFs[0]
+  if (xosanNetwork.$id !== firstVif.$network.$id) {
+    console.log('VIF in wrong network (' + firstVif.$network.name_label + '), moving to correct one: ' + xosanNetwork.name_label)
```
julien-f: `console.warn()`.

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+  await xapi.call('VM.set_xenstore_data', vm.$ref, params.xenstore_data)
+  if (increaseDataDisk) {
+    const dataDisk = vm.$VBDs.map(vbd => vbd.$VDI).find(vdi => vdi && vdi.name_label === 'xosan_data')
+    const srFreeSpace = sr.physical_size - sr.physical_utilisation
+    // we use a percentage because it looks like the VDI overhead is proportional
+    const newSize = trucate2048((srFreeSpace + dataDisk.virtual_size) * XOSAN_DATA_DISK_USEAGE_RATIO)
+    await xapi._resizeVdi(dataDisk, newSize)
+  }
+  await xapi.startVm(vm)
+  console.log('waiting for boot of ', ip)
```
julien-f: No `console.log()`, you can use `debug` if you want to.

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+  // wait until we find the assigned IP in the networks, we are just checking the boot is complete
+  const vmIsUp = vm => Boolean(vm.$guest_metrics && Object.values(vm.$guest_metrics.networks).find(value => value === ip))
```
julien-f: Use `lodash/includes`.

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+  $onFailure(async () => await xapi.deleteNetwork(xosanNetwork)::pCatch(noop))
+  await Promise.all(xosanNetwork.$PIFs.map(pif => setPifIp(xapi, pif, NETWORK_PREFIX + (hostIpLastNumber++))))
+
+  return xosanNetwork
+})
+async function getOrCreateSshKey (xapi) {
+  let sshKey = xapi.xo.getData(xapi.pool, 'xosan_ssh_key')
+  if (!sshKey) {
+    try {
+      await fs.access(SSH_KEY_FILE, fs.constants.R_OK)
```
julien-f: Avoid [testing access, simply read and react to errors](https://en.wikipedia.org/wiki/Time_of_check_to_time_of_use).

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+      creation: 'replica ' + redundancy + ' ',
+      extra: ['gluster volume set xosan cluster.data-self-heal on']
+    },
+    disperse: {
+      creation: 'disperse ' + ipAndHosts.length + ' redundancy ' + redundancy + ' ',
+      extra: []
+    }
+  }
+  for (let i = 1; i < ipAndHosts.length; i++) {
+    await remoteSsh(xapi, firstIpAndHost, 'gluster peer probe ' + ipAndHosts[i].address)
```
julien-f: Why not doing it in parallel? (`map` + `Promise.all()`)

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+  }
+  let brickVms = ipAndHosts
+  let creation = configByType[glusterType].creation
+  if (arbiter) {
+    await remoteSsh(xapi, firstIpAndHost, 'gluster peer probe ' + arbiter.address)
+    brickVms = brickVms.concat([arbiter])
```
julien-f: You can use ES2015 syntax if you prefer: `brickVms = [ ...brickVms, arbiter ]`.

And in fact, because you do not need immutable values, you can simply `brickVms.push(arbiter)`.

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+  await remoteSsh(xapi, firstIpAndHost, 'gluster volume start xosan')
+}
+
+export const createSR = defer.onFailure(async function ($onFailure, { template, pif, vlan, srs, glusterType, redundancy }) {
+  if (!this.requestResource) {
+    throw new Error('requestResource is not a function')
+  }
+
+  if (srs.length < 1) {
+    return // TODO: throw an error
```
julien-f: Why `TODO`?

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+  }
+
+  let vmIpLastNumber = 101
```
julien-f: Magical number?

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+  let xapi = this.getXapi(srs[0])
+  if (CURRENTLY_CREATING_SRS[xapi.pool.$id]) {
+    throw new Error('createSR is already running for this pool')
+  } else {
```
julien-f: `else` is unnecessary here.

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+  if (!this.requestResource) {
+    throw new Error('requestResource is not a function')
+  }
+
+  if (srs.length < 1) {
+    return // TODO: throw an error
+  }
+
+  let vmIpLastNumber = 101
+  let xapi = this.getXapi(srs[0])
```
julien-f: Always use `const` unless you need to reassign.

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+  if (CURRENTLY_CREATING_SRS[xapi.pool.$id]) {
+    throw new Error('createSR is already running for this pool')
+  } else {
+    CURRENTLY_CREATING_SRS[xapi.pool.$id] = true
+  }
+  try {
+    let xosanNetwork = await createNetworkAndInsertHosts(xapi, pif, vlan)
+    $onFailure(async () => await xapi.deleteNetwork(xosanNetwork)::pCatch(noop))
```
julien-f: `async` + `return await` is usually unnecessary, simply returns the promise :)

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+    let sshKey = await getOrCreateSshKey(xapi)
+    const srsObjects = map(srs, srId => xapi.getObject(srId))
+
+    const vmParameters = map(srs, srId => {
+      const sr = xapi.getObject(srId)
+      const host = xapi.getObject(xapi.getObject(sr.$PBDs[0]).host)
```
julien-f: Use either the magical props (preferred) or `xapi.getObject()`, no need to use both:

```js
const host = sr.$PBDs[0].$host
```

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+      return {
+        sr,
+        host,
+        name_label: 'XOSAN - ' + sr.name_label + ' - ' + host.name_label,
```
julien-f: Use ES2015 strings: `` `XOSAN - ${sr.name_label} - ${host.name_label}` ``.

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+  { layout: 'disperse', redundancy: 4, capacity: 8 },
+  { layout: 'replica', redundancy: 2, capacity: 6 }]
+POSSIBLE_CONFIGURATIONS[13] = [{ layout: 'disperse', redundancy: 5, capacity: 8 }]
+POSSIBLE_CONFIGURATIONS[14] = [
+  { layout: 'disperse', redundancy: 6, capacity: 8 },
+  { layout: 'replica', redundancy: 2, capacity: 7 }]
+POSSIBLE_CONFIGURATIONS[15] = [
+  { layout: 'disperse', redundancy: 7, capacity: 8 },
+  { layout: 'replica', redundancy: 3, capacity: 5 }]
+POSSIBLE_CONFIGURATIONS[16] = [{ layout: 'replica', redundancy: 2, capacity: 8 }]
```
julien-f: Either define `POSSIBLE_CONFIGURATIONS` using an array literal, or use an object if you need to see the keys.

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+  if (!configurations) {
+    return null
+  }
+  if (count > 0) {
+    let xapi = this.getXapi(lvmSrs[0])
+    let srs = map(lvmSrs, srId => xapi.getObject(srId))
+    let srSizes = map(srs, sr => sr.physical_size - sr.physical_utilisation)
+    let minSize = Math.min.apply(null, srSizes)
+    let brickSize = (minSize - XOSAN_VM_SYSTEM_DISK_SIZE) * XOSAN_DATA_DISK_USEAGE_RATIO
+    return configurations.map(conf => Object.assign({}, conf, { availableSpace: brickSize * conf.capacity }))
```
julien-f:
```js
{ ...conf, availableSpace: brickSize * conf.capacity }
```

src/xapi-object-to-xo.js
```diff
@@ -386,6 +386,8 @@ const TRANSFORMS = {
       tags: obj.tags,
       usage: +obj.virtual_allocation,
       VDIs: link(obj, 'VDIs'),
+      other_config: obj.other_config,
```
julien-f: Exposing `other_config` is usually a bad practice because it breaks the encapsulation of XO objects.

I also do it sometimes because it's easier, but if there is an easy way to avoid it (like exposing a single value), it should be preferred.

src/api/plugin.js
```diff
@@ -123,3 +123,33 @@ test.params = {
...
+
+registerResource.description = 'Register a resource via cloud plugin'
+
+registerResource.params = {
+  namespace: {
+    type: 'string'
+  }
+}
+
+registerResource.permission = 'admin'
```
julien-f: Those methods should be exposed by the plugin itself: `xo.addApiMethod()`.

src/api/xosan.js
```diff
@@ -0,0 +1,493 @@
...
+      stderr += buffer.toString()
+    })
+    child.on('close', (code) => {
+      if (code !== 0) {
+        reject(stderr)
+      } else {
+        resolve(resp)
+      }
+    })
+  })
```
julien-f: Should be easy to fix (https://github.com/sindresorhus/execa).

src/api/xosan.js
nraynaud: I don't think it's the right place for that thinking here. Detecting the error means parsing ssh's stdout.

src/api/xosan.js
```diff
@@ -0,0 +1,493 @@
...
+   performance.quick-read: off
+   performance.stat-prefetch: on
+   performance.strict-write-ordering: off
+   cluster.server-quorum-type: server
+   cluster.quorum-type: auto
+   */
+  const info = splitLines(fromPairs(result.trim())
+    .map(line => line.split(':').map(val => val.trim()))
+    // some lines have more than one ':', re-assemble them
+    .map(line => [line[0], line.slice(1).join(':')]))
```
julien-f: Should be easy to fix as well (https://github.com/vatesfr/xo-server/pull/505#discussion_r102682973).

src/api/xosan.js
```diff
@@ -0,0 +1,493 @@
...
+  const getNumber = item => +item.substr(5)
```
julien-f: Inline it :)

src/api/xosan.js
```diff
@@ -0,0 +1,493 @@
...
+  const brickKeys = Object.keys(info).filter(key => key.match(/^Brick[1-9]/)).sort((i1, i2) => getNumber(i1) - getNumber(i2))
```
julien-f: Use `mapFilter` :)

```js
const brickKeys = mapFilter(info, (_, key) => {
  const matches = /^brick([0-9]+)/i.exec(key)
  return matches && matches[1]
}).sort((i, j) => +i - +j)
```

src/api/xosan.js
julien-f: You can change it.

src/api/xosan.js
```diff
@@ -0,0 +1,493 @@
...
+    const dataDisk = vm.$VBDs.map(vbd => vbd.$VDI).find(vdi => vdi && vdi.name_label === 'xosan_data')
+    const srFreeSpace = sr.physical_size - sr.physical_utilisation
+    // we use a percentage because it looks like the VDI overhead is proportional
+    const newSize = trucate2048((srFreeSpace + dataDisk.virtual_size) * XOSAN_DATA_DISK_USEAGE_RATIO)
+    await xapi._resizeVdi(dataDisk, newSize)
+  }
+  await xapi.startVm(vm)
+  debug('waiting for boot of ', ip)
+  // wait until we find the assigned IP in the networks, we are just checking the boot is complete
+  const vmIsUp = vm => Boolean(vm.$guest_metrics && includes(Object.values(vm.$guest_metrics.networks), ip))
```
julien-f: `Object.values()` is unnecessary.

src/api/xosan.js
```diff
@@ -0,0 +1,493 @@
...
+  $onFailure(async () => await xapi.deleteNetwork(xosanNetwork)::pCatch(noop))
+  await Promise.all(xosanNetwork.$PIFs.map(pif => setPifIp(xapi, pif, NETWORK_PREFIX + (hostIpLastNumber++))))
+
+  return xosanNetwork
+})
+async function getOrCreateSshKey (xapi) {
+  let sshKey = xapi.xo.getData(xapi.pool, 'xosan_ssh_key')
+  if (!sshKey) {
+    try {
+      await fs.access(SSH_KEY_FILE, fs.constants.R_OK)
```
julien-f: Easy to fix as well.

src/api/xosan.js
```diff
@@ -0,0 +1,493 @@
...
+      extra: ['gluster volume set xosan cluster.data-self-heal on']
+    },
+    disperse: {
+      creation: 'disperse ' + ipAndHosts.length + ' redundancy ' + redundancy + ' ',
+      extra: []
+    }
+  }
+  await Promise.all(
+    map(ipAndHosts, ({ address }) => remoteSsh(xapi, firstIpAndHost, 'gluster peer probe ' + address))
+  )
```
julien-f: Revert this change, we'll need @nraynaud input on this.

src/api/xosan.js
```diff
@@ -0,0 +1,493 @@
...
+const createNetworkAndInsertHosts = defer.onFailure(async function ($onFailure, xapi, pif, vlan) {
+  let hostIpLastNumber = 1
+  const xosanNetwork = await xapi.createNetwork({
+    name: 'XOSAN network',
+    description: 'XOSAN network',
+    pifId: pif._xapiId,
+    mtu: 9000,
+    vlan: +vlan
+  })
+  $onFailure(async () => await xapi.deleteNetwork(xosanNetwork)::pCatch(noop))
```
julien-f: Remove all `async () => await`.

src/api/xosan.js
julien-f: We'll need @nraynaud input on this.

src/api/xosan.js
julien-f: Nope, simply detecting `ENOENT` errors of `fs.readFile()` :)

nraynaud: you are right, that check as traveled a bit.

src/api/xosan.js
```diff
@@ -0,0 +1,484 @@
...
+  })
+  await new Promise((resolve, reject) => arp.table((err, entry) => {
+    if (entry) {
+      const brick = info['Bricks'].find(element => element.config.split(':')[0] === entry.ip)
+      if (brick) {
+        brick.mac = entry.mac
+      }
+    }
+    if (!entry && !err) {
+      resolve()
```
pdonias: @nraynaud What if we never get in the `if`?

src/api/xosan.js
nraynaud: it's just an object whose keys are small ints, I don't know what to answer. Either you want to cast them to string in an object, or you make an array initializer but you can't read the indexes when you revise the code. I didn't really care that the array would become sparse.

julien-f: @pdonias Let's go for an object then, the casts will be implicit.

src/api/xosan.js
nraynaud: maybe reject()?

src/api/xosan.js
```diff
@@ -87,28 +76,27 @@ export async function getVolumeInfo ({ sr }) {
...
-    .map(line => [line[0], line.slice(1).join(':')]))
-  let getNumber = item => parseInt(item.substr(5))
-  let brickKeys = Object.keys(info).filter(key => key.match(/^Brick[1-9]/)).sort((i1, i2) => getNumber(i1) - getNumber(i2))
+  const info = splitLines(fromPairs(result.trim()))
+    .map(line => splitFirst(line, ':').map(val => val.trim()))
+
+  const brickKeys = mapFilter(info, (_, key) => {
+    const matches = /^brick([0-9]+)/i.exec(key)
+    return matches && matches[1]
+  }).sort((i, j) => +i - +j)
```
julien-f: We broke the code with this change, let's revert (but use `lodash.filter` instead of `Object.keys().filter()`).

src/api/xosan.js
```diff
@@ -87,28 +76,27 @@ export async function getVolumeInfo ({ sr }) {
...
+
   // expected brickKeys : [ 'Brick1', 'Brick2', 'Brick3' ]
   info['Bricks'] = brickKeys.map(key => {
     const ip = info[key].split(':')[0]
     return { config: info[key], ip: ip, vm: giantIPtoVMDict[ip] }
   })
-  await new Promise((resolve, reject) => arp.table((err, entry) => {
+  arp.table((_, entry) => {
```
julien-f: I think there was an error, because the code would get stuck `if (entry)`.

```js
const entry = await pFromCallback(cb => arp.table(cb))
if (entry) {
  // ...
}
```

src/api/xosan.js
nraynaud: I was not feeling good about having concurrent peering on the same machine. But I have no proof it's an issue. I couldn't find a command to peer more than one pair of machines at a time. The same for the huge list of configuration commands later in the file, these are parts of the system whose resilience to concurrency I was not willing to test.

src/api/xosan.js
```diff
@@ -0,0 +1,479 @@
...
+}
+
+async function prepareGlusterVm (xapi, vmAndParam, xosanNetwork, increaseDataDisk = true) {
+  let vm = vmAndParam.vm
+  // refresh the object so that sizes are correct
+  const params = vmAndParam.params
+  const ip = params.xenstore_data['vm-data/ip']
+  const sr = xapi.getObject(params.sr.$id)
+  await xapi._waitObjectState(sr.$id, sr => Boolean(sr.$PBDs))
+  const host = xapi.getObject(xapi.getObject(sr.$PBDs[0]).host)
```
julien-f: Use magical properties.

src/api/xosan.js
```diff
@@ -79,24 +80,22 @@ export async function getVolumeInfo ({ sr }) {
   const info = splitLines(fromPairs(result.trim()))
     .map(line => splitFirst(line, ':').map(val => val.trim()))
 
-  const brickKeys = mapFilter(info, (_, key) => {
-    const matches = /^brick([0-9]+)/i.exec(key)
-    return matches && matches[1]
-  }).sort((i, j) => +i - +j)
+  let getNumber = item => parseInt(item.substr(5))
+  let brickKeys = filter(info, (_, key) => key.match(/^Brick[1-9]/)).sort((i1, i2) => getNumber(i1) - getNumber(i2))
```
julien-f: `const`.

src/api/xosan.js
```diff
@@ -79,24 +80,22 @@ export async function getVolumeInfo ({ sr }) {
   const info = splitLines(fromPairs(result.trim()))
     .map(line => splitFirst(line, ':').map(val => val.trim()))
 
-  const brickKeys = mapFilter(info, (_, key) => {
-    const matches = /^brick([0-9]+)/i.exec(key)
-    return matches && matches[1]
-  }).sort((i, j) => +i - +j)
+  let getNumber = item => parseInt(item.substr(5))
```
julien-f: `parseInt` → `+`.

src/api/xosan.js
```diff
@@ -126,7 +125,7 @@ async function prepareGlusterVm (xapi, vmAndParam, xosanNetwork, increaseDataDis
   const ip = params.xenstore_data['vm-data/ip']
   const sr = xapi.getObject(params.sr.$id)
   await xapi._waitObjectState(sr.$id, sr => Boolean(sr.$PBDs))
-  const host = xapi.getObject(xapi.getObject(sr.$PBDs[0]).host)
+  const host = xapi.getObject(sr.$PBDs[0]).$host
```
julien-f:
```js
const host = sr.$PBDs[0].$host
```

src/api/xosan.js
```diff
@@ -0,0 +1,478 @@
...
+    const ip = info[key].split(':')[0]
+    return { config: info[key], ip: ip, vm: giantIPtoVMDict[ip] }
+  })
+  const entry = await pFromCallback(cb => arp.table(cb))
+  if (entry) {
+    const brick = info['Bricks'].find(element => element.config.split(':')[0] === entry.ip)
+    if (brick) {
+      brick.mac = entry.mac
+    }
+  }
```
julien-f: @nraynaud Does that makes sense to you?

It will:
- throw in case of error
- enter the if only if `entry`

nraynaud: Yes, that looks good.
Are we still using the mac address at all? It was made to find the corresponding virtual machine, depending on the UI changes, you could scrap this entire piece of code.

julien-f: No idea, we'll keep it for this first release :p

src/api/xosan.js
```diff
@@ -87,28 +77,25 @@ export async function getVolumeInfo ({ sr }) {
...
    cluster.quorum-type: auto
    */
-  let info = fromPairs(result.trim().split('\n')
-    .map(line => line.split(':').map(val => val.trim()))
-    // some lines have more than one ':', re-assemble them
-    .map(line => [line[0], line.slice(1).join(':')]))
-  let getNumber = item => parseInt(item.substr(5))
-  let brickKeys = Object.keys(info).filter(key => key.match(/^Brick[1-9]/)).sort((i1, i2) => getNumber(i1) - getNumber(i2))
+  const info = splitLines(fromPairs(result.trim()))
+    .map(line => splitFirst(line, ':').map(val => val.trim()))
```
julien-f:
```js
const info = fromPairs(
  splitLines(result.trim()).map(line =>
    splitFirst(line, ':').map(val => val.trim())
  )
)
```

PR #507, opened 2017-02-24
feat(prepareGlusterVm): use reported size in case of failure

julien-f:


julien-f:
I wanted to but it's still in use https://github.com/vatesfr/xo-server/blob/0f85826ab0db9ee9c863716fa2eaf3e9435391e2/src/api/xosan.js#L449

PR #514, opened 2017-03-03
fix(VM.resume): adding resumeVm() in XAPI

badrAZ:


src/api/vm.coffee
```diff
@@ -920,18 +920,13 @@ exports.suspend = suspend
...
 #---------------------------------------------------------------------
 
-resume = $coroutine ({vm, force}) ->
-  # FIXME: WTF this is?
-  if not force
-    force = true
-
-  yield @getXapi(vm).call 'VM.resume', vm._xapiRef, false, force
+resume = $coroutine ({vm}) ->
+  yield @getXapi(vm).resumeVm(vm._xapiRef)
```
julien-f: You can remove `$coroutine`, and returns the promise directly :)

src/xapi/mixins/vm.js
```diff
@@ -331,7 +331,7 @@ export default {
     tags: true
   }),
 
-  async editVm (id, props, checkLimits) {
+  editVm (id, props, checkLimits) {
```
julien-f: Revert this change, it's not related to the current PR.

src/xapi/mixins/vm.js
```diff
@@ -341,5 +341,10 @@ export default {
       await this._snapshotVm(snapshot.$snapshot_of)
     }
     return this.call('VM.revert', snapshot.$ref)
+  },
+
+  resumeVm (ref) {
```
julien-f: Don't expect the param to be a ref, expect it to be an identifier (`vmId`), fetch the ref through the corresponding object: `this.getObject(vmId).$ref`.

src/xapi/mixins/vm.js
```diff
@@ -341,5 +341,10 @@ export default {
...
+    // the force parameter is always true
+    return this.call('VM.resume', ref, false, true)
```
julien-f: This is a public method, make sure the returned (async or not) value is always the same.

PR #516, opened 2017-03-07
fix(Xapi#createVm): disks should be bootable according to the `vdis` param

pdonias:
Fixes vatesfr/xo-web#1994


src/xapi/mixins/vm.js
```diff
@@ -158,7 +158,7 @@ export default {
           .then(ref => this._getOrWaitObject(ref))
           .then(vdi => this._createVbd(vm, vdi, {
             // Only the first VBD if installMethod is not cd is bootable.
```
julien-f: Maybe remove the comment if no longer relevant.

julien-f:
@olivierlambert Be careful that this PR reverts a previous change.

I don't remember why we did this but there should have been a reason…

olivierlambert:
@julien-f copy that, I'll make extensive tests

olivierlambert:
### First problem

This **breaks all PV guest creation** (PV+ISO). It seems that the CD VBD is created with a "bootable" at true, which is NOT something that should be done. In PV, only the first disk should be bootable at true and that's it.

The [user reported originally](https://xen-orchestra.com/forum/topic/427/self-service-secondary-disk-bug) that extra disk were having the "bootable" flag at true, which should be avoided. IDK how to trigger this.

IMHO, the original code is correct, not this PR. In the end, it will be probably wise to hide the "bootable" checkbox. @pdonias is this possible the behavior is different when creating a VM on self service?

### Second problem

This seems to fail if I add a new disk, and let "bootable" not checked. Then, no bootable value is passed for this VDI and it VM creation fails with the error below.

Workaround: check and uncheck the box to make it work. I assume it's an UI issue.

```
vm.create
{
  "clone": true,
  "existingDisks": {},
  "installation": {
    "method": "cdrom",
    "repository": "f8329dbc-16b0-4f7e-b061-434ec7b517a0"
  },
  "name_label": "testOLT",
  "template": "4f4f3a4f-2529-6c0e-9404-717f6048796a",
  "VDIs": [
    {
      "device": "1",
      "size": 8589934592,
      "bootable": true,
      "type": "system",
      "SR": "e6616eb6-d719-b7a3-95a8-73881b90c25e",
      "name_description": "Created by XO",
      "name_label": "Debian Wheezy 7.0 (64-bit)_1"
    },
    {
      "device": "2",
      "name_description": "Created by XO",
      "name_label": "testOLT_2",
      "SR": "e6616eb6-d719-b7a3-95a8-73881b90c25e",
      "type": "system",
      "size": 2147483648
    }
  ],
  "VIFs": [
    {
      "id": 0,
      "network": "887b96dc-1e24-b616-1523-9c5818c534eb",
      "allowedIpv4Addresses": [],
      "allowedIpv6Addresses": []
    }
  ],
  "CPUs": 1,
  "cpuWeight": null,
  "cpuCap": null,
  "name_description": "Template that allows VM installation from Xen-aware Debian-based distros. To use this template from the CLI, install your VM using vm-install, then set other-config-install-repository to the path to your network repository, e.g. http://<server>/<path>",
  "memoryMax": 268435456,
  "bootAfterCreate": true,
  "coreOs": false,
  "tags": []
}
{
  "message": "invalid parameters",
  "stack": "XoError: invalid parameters
    at Object.factory [as invalidParameters] (/home/olivier/xo-server/node_modules/xo-common/src/api-errors.js:21:31)
    at Xo.checkParams (/home/olivier/xo-server/dist/xo-mixins/xo-mixins/api.js:67:17)
    at Api._callee2$ (/home/olivier/xo-server/dist/xo-mixins/xo-mixins/api.js:255:18)
    at tryCatch (/home/olivier/xo-server/node_modules/regenerator-runtime/runtime.js:64:40)
    at GeneratorFunctionPrototype.invoke [as _invoke] (/home/olivier/xo-server/node_modules/regenerator-runtime/runtime.js:299:22)
    at GeneratorFunctionPrototype.prototype.(anonymous function) [as next] (/home/olivier/xo-server/node_modules/regenerator-runtime/runtime.js:116:21)
    at step (/home/olivier/xo-server/node_modules/babel-runtime/helpers/asyncToGenerator.js:17:30)
    at /home/olivier/xo-server/node_modules/babel-runtime/helpers/asyncToGenerator.js:28:13
    at run (/home/olivier/xo-server/node_modules/core-js/library/modules/es6.promise.js:87:22)
    at /home/olivier/xo-server/node_modules/core-js/library/modules/es6.promise.js:100:28
    at flush (/home/olivier/xo-server/node_modules/core-js/library/modules/_microtask.js:18:9)
    at _combinedTickCallback (internal/process/next_tick.js:67:7)
    at process._tickCallback (internal/process/next_tick.js:98:9)",
  "code": 10,
  "data": {
    "errors": [
      {
        "code": null,
        "reason": "optional",
        "message": "is missing and not optional",
        "property": "@.VDIs[1].bootable"
      }
    ]
  }
}
```

olivierlambert:
Okay I think I get it: our code works for a new PV guest without existing disk. But if we create a PV from a template with disk, and add one disk (or more), the first extra will be flagged as bootable:true.

But this is a mistake, because the existing disk is already set a bootable. So it fails to boot.

src/xapi/mixins/vm.js
```diff
@@ -143,6 +144,7 @@ export default {
     //
     // TODO: set vm.suspend_SR
     if (!isEmpty(vdis)) {
+      const nBootable = filter(vdis, 'bootable').length
```
julien-f: Use `sumBy` or `reduce`.

julien-f:
IMHO, the correct approach is:

- do not let the user specify which disks are bootable (update xo-web)
- if there are no bootable VBDs, set the first disk as bootable

src/xapi/mixins/vm.js
```diff
@@ -143,6 +143,7 @@ export default {
     //
     // TODO: set vm.suspend_SR
     if (!isEmpty(vdis)) {
+      const bootableExistingDisk = find(existingVdis, 'bootable')
```
julien-f: `let hasBootableDisk = !!find(vm.VBDs, 'bootable')` and change when you create a CD.

src/xapi/mixins/vm.js
```diff
@@ -159,7 +160,7 @@ export default {
           .then(ref => this._getOrWaitObject(ref))
           .then(vdi => this._createVbd(vm, vdi, {
             // Either the CD or the 1st disk is bootable (only useful for PV VMs)
-            bootable: installMethod !== 'cd' && !bootableExistingDisk ? !i : false,
+            bootable: hasBootableDisk ? false : !i,
```
julien-f: `!(hasBootable || i)`

PR #521, opened 2017-03-24
feat(vm): affinity host

pdonias:
See vatesfr/xo-web#1983


src/xapi/mixins/vm.js
```diff
@@ -229,6 +230,17 @@ export default {
   //
   // Params do not correspond directly to XAPI props.
   _editVm: makeEditObject({
+    affinityHost: {
+      get: 'affinity',
+      set (value, vm) {
+        this._setObjectProperty(
```
julien-f: You should return the promise.

PR #522, opened 2017-03-24
feat(jobs): configure job timeout

badrAZ:
Fixes vatesfr/xo-web#1956


src/job-executor.js
```diff
@@ -164,30 +165,33 @@ export default class JobExecutor {
...
-            error: serializeError(reason)
-          })
-
-          call.error = reason
-          call.end = Date.now()
-        }
-      )
+      return (job.timeout
+        ? this.xo.callApiMethod(connection, job.method, assign({}, params))::timeout(job.timeout * 1e3)
+        : this.xo.callApiMethod(connection, job.method, assign({}, params)))
```
julien-f: Use an intermediary variable, and avoid code copy.

PR #526, opened 2017-04-06
feat: restore the VM power state at snapshot after reverting to a snapshot

badrAZ:
Fixes vatesfr/xo-web#834


src/xapi/mixins/vm.js
```diff
@@ -353,7 +353,16 @@ export default {
...
       await this._snapshotVm(snapshot.$snapshot_of)
     }
-    return this.call('VM.revert', snapshot.$ref)
+    await this.call('VM.revert', snapshot.$ref)
+    if (snapshot.snapshot_info['power-state-at-snapshot'] === 'Running') {
+      const vm = this.getObject(snapshot.snapshot_of)
+      if (vm.power_state === 'Halted') {
+        await this.startVm(vm.uuid)
+      }
+      if (vm.power_state === 'Suspended') {
```
julien-f: `else if`, the VM cannot be in both states at the same time.

src/xapi/mixins/vm.js
```diff
@@ -353,7 +353,16 @@ export default {
     if (snapshotBefore) {
       await this._snapshotVm(snapshot.$snapshot_of)
     }
-    return this.call('VM.revert', snapshot.$ref)
+    await this.call('VM.revert', snapshot.$ref)
+    if (snapshot.snapshot_info['power-state-at-snapshot'] === 'Running') {
+      const vm = this.getObject(snapshot.snapshot_of)
```
julien-f: `const vm = snapshot.$snapshot_of`.

src/xapi/mixins/vm.js
```diff
@@ -353,7 +353,16 @@ export default {
...
+      if (vm.power_state === 'Halted') {
+        await this.startVm(vm.uuid)
+      }
+      if (vm.power_state === 'Suspended') {
+        await this.resumeVm(vm.uuid)
+      }
```
julien-f: Should this method waits for the new VM to start/resume?

Should this method be reported as failed if the new VM failed to start/resume?

/cc @olivierlambert

src/xapi/mixins/vm.js
```diff
@@ -353,7 +353,15 @@ export default {
...
     }
-    return this.call('VM.revert', snapshot.$ref)
+    await this.call('VM.revert', snapshot.$ref)
+    if (snapshot.snapshot_info['power-state-at-snapshot'] === 'Running') {
+      const vm = snapshot.$snapshot_of
+      if (vm.power_state === 'Halted') {
+        await this.startVm(vm.uuid)
+      } else if (vm.power_state === 'Suspended') {
+        await this.resumeVm(vm.uuid)
+      }
```
julien-f: Let's do like XenCenter: start/resume async and ignore any errors.

PR #533, opened 2017-04-24
fix(Xo#registerUser): only accept users without passwords

julien-f:


src/xo-mixins/authentication.js
```diff
@@ -91,26 +93,25 @@ export default class {
...
+        }
+
+        assert(typeof result.username === 'string')
+        return this._xo.registerUser(undefined, result.username)
       } catch (error) {
         // DEPRECATED: Authentication providers may just throw `null`
         // to indicate they could not authenticate the user without
         // any special errors.
-        if (error) console.error(error.stack || error)
+        if (error !== null) console.warn(error.stack || error)
```
pdonias: Can `error` be undefined?

julien-f: Should not happen, I accept a failure in that case ;)

PR #536, opened 2017-04-25
chore(package): remove babel-preset-stage-0

julien-f:
Depends on the necessary transforms directly, it should help debugging with Node > 7.5.0.


julien-f:
Issue with Node < 7.5.0 for async methods (unsupported feature) using `super` (supported feature):

```js
class B extends A {
  async foo () {
    return await super.foo()
  }
}
```

julien-f:
To merge once babel/babel#3930 is fixed.

PR #537, opened 2017-05-02
feat(vm): ability to choose the cores per socket when creating or editing a VM

badrAZ:
Fixes vatesfr/xo-web#130


src/xapi-object-to-xo.js
```diff
@@ -79,7 +79,11 @@ const TRANSFORMS = {
       tags: obj.tags,
       name_description: obj.name_description,
       name_label: obj.name_label || obj.$master.name_label,
-      xosanPackInstallationTime: toTimestamp(obj.other_config.xosan_pack_installation_time)
+      xosanPackInstallationTime: toTimestamp(obj.other_config.xosan_pack_installation_time),
+      cpus: {
+        cores: +obj.cpu_info.cpu_count,
+        sockets: +obj.cpu_info.socket_count
```
julien-f: Are we sure `obj.cpu_info.{cpu,socket}_count` are always defined?

On XS 6.2, 6.5, 7.0, 7.1?

src/xapi-object-to-xo.js
```diff
@@ -367,6 +371,8 @@ const TRANSFORMS = {
       } = obj.platform)
     }
 
+    vm.coresPerSocket = obj.platform['cores-per-socket']
```
julien-f: Shouldn't it be a number?

src/xapi/mixins/vm.js
```diff
@@ -259,6 +259,12 @@ export default {
       }
     },
 
+    coresPerSocket: {
+      set (core, vm) {
```
julien-f: `core` → `coresPerSocket` or `n`.

src/xapi-object-to-xo.js
badrAZ: It is only defined in 7.0, 7.1,  i put 4 cores as default value?

src/xapi-object-to-xo.js
badrAZ: Yes, it should. I will cast it.

src/xapi-object-to-xo.js
julien-f: Does it makes sense to have default values?

If not, simply set this values to `undefined`.

src/xapi-object-to-xo.js
julien-f: Be careful to not cast `undefined`.

src/xapi-object-to-xo.js
```diff
@@ -80,7 +80,7 @@ const TRANSFORMS = {
       name_description: obj.name_description,
       name_label: obj.name_label || obj.$master.name_label,
       xosanPackInstallationTime: toTimestamp(obj.other_config.xosan_pack_installation_time),
-      cpus: {
+      cpus: obj.cpu_info && {
```
julien-f: I would prefer avoiding conditional nested props.

src/xapi-object-to-xo.js
```diff
@@ -371,7 +371,7 @@ const TRANSFORMS = {
       } = obj.platform)
     }
 
-    vm.coresPerSocket = obj.platform['cores-per-socket']
+    vm.coresPerSocket = obj.platform['cores-per-socket'] && +obj.platform['cores-per-socket']
```
julien-f: Use an intermediary variable.

src/xapi-object-to-xo.js
julien-f: Use an intermediary variable.

PR #538, opened 2017-05-03
feat(xo-server-recover-account): reset password and set permission to admin

julien-f:
Fixes vatesfr/xo-web#2129


julien-f:
Maybe it should be called `xo-server-recover-admin` instead?

src/recover-account-cli.js
```diff
@@ -0,0 +1,38 @@
...
+  if (
+    name === undefined ||
+    name === '--help' ||
+    name === '-h'
+  ) {
+    return `
+xo-server-recovery-account <user name or email>
+
+    If the user does not exist, it is created, if it exists, updates
+    its password and reset its permission to administrator.
```
pdonias: resets*

julien-f: Thx.

PR #544, opened 2017-05-11
fix: deferreds should not throw errors

julien-f:


julien-f:
@pdonias Could you review this? I'd like these changes to ship this afternoon.

PR #545, opened 2017-05-12
feat(servers): allow unauthorized certificates

badrAZ:
Fixes vatesfr/xo-web#2138


src/api/server.js
```diff
@@ -9,9 +9,10 @@ export async function add ({
   username,
   password,
   readOnly,
-  autoConnect = true
+  autoConnect = true,
+  allowUnauthorized
 }) {
-  const server = await this.registerXenServer({label, host, username, password, readOnly})
+  const server = await this.registerXenServer({label, host, username, password, readOnly, allowUnauthorized})
```
julien-f: It would be much easier to use rest properties syntax in the destructuring, and pass the object directly :)

src/api/server.js
```diff
@@ -75,8 +80,8 @@ getAll.permission = 'admin'
 
 // -------------------------------------------------------------------
 
-export async function set ({id, label, host, username, password, readOnly}) {
-  await this.updateXenServer(id, {label, host, username, password, readOnly})
+export async function set ({id, label, host, username, password, readOnly, allowUnauthorized}) {
+  await this.updateXenServer(id, {label, host, username, password, readOnly, allowUnauthorized})
```
julien-f: Use rest properties syntax.

src/models/server.js
```diff
@@ -15,12 +15,12 @@ export class Servers extends Collection {
     return Server
   }
 
-  async create ({label, host, username, password, readOnly}) {
+  async create ({label, host, username, password, readOnly, allowUnauthorized}) {
```
julien-f: No need to destructure.

PR #546, opened 2017-05-12
feat(http-request): always use HTTP proxy

julien-f:


src/xapi/mixins/patching.js
```diff
@@ -19,17 +17,15 @@ import {
...
 } from '../utils'
 
 export default {
   // FIXME: should be static
   @debounce(24 * 60 * 60 * 1000)
   async _getXenUpdates () {
-    const { readAll, statusCode } = await httpRequest(
-      'http://updates.xensource.com/XenServer/updates.xml',
-      { agent: httpProxy }
+    const { readAll, statusCode } = await this._xo.httpRequest(
```
julien-f: `this._xo` is `undefined`.

src/utils.js
```diff
@@ -226,6 +226,19 @@ export function extractProperty (obj, prop) {
 
 // -------------------------------------------------------------------
 
+// Returns the first defined (non-null, non-undefined) value.
+export const firstDefined = function () {
+  const n = arguments.length
+  for (let i = 0; i < n; ++i) {
+    const arg = arguments[i]
+    if (arg !== undefined) {
```
julien-f: This code comes from xo-web, what should be the proper behavior according to you?

pdonias: In xo-web, it must be kept like this since we often use it with `myFunction(arg1, arg2, null)` to get `null` as the fallback value. Here, both would work so we can keep it identical to xo-web. But anyway, the comments should not say "non-null".

julien-f:
To test:

- [ ] CR
- [ ] Delta Backup
- [ ] Patching

PR #547, opened 2017-05-16
feat(job-executor/execStatus): inject timezone from related schedule

badrAZ:
See vatesfr/xo-web#2133

### DIFF
--- a/src/job-executor.js
+++ b/src/job-executor.js
@@ -147,13 +147,13 @@ export default class JobExecutor {
 
     connection.set('user_id', job.userId)
 
-    const { timezone } = find(await this.xo.getAllSchedules(), ['job', job.id])
+    const schedule = find(await this.xo.getAllSchedules(), { job: job.id })
 
     const execStatus = {
+      calls: {},
       runJobId,
       start: Date.now(),
-      timezone,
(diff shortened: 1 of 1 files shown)


src/job-executor.js
```diff
@@ -146,9 +147,12 @@ export default class JobExecutor {
 
     connection.set('user_id', job.userId)
 
+    const { timezone } = find(await this.xo.getAllSchedules(), ['job', job.id])
+
     const execStatus = {
       runJobId,
       start: Date.now(),
+      timezone,
       calls: {}
```
julien-f: Sort properties.

src/job-executor.js
```diff
@@ -146,9 +147,12 @@ export default class JobExecutor {
 
     connection.set('user_id', job.userId)
 
+    const { timezone } = find(await this.xo.getAllSchedules(), ['job', job.id])
```
julien-f: What if there are no schedules for this job?

julien-f: Use `{ job: job.id }` as predicate instead of the array form.

badrAZ: It's an error, i will fix it.

PR #548, opened 2017-05-17
feat(backups): transferred data size for CR & DR

badrAZ:


src/xapi/index.js
```diff
@@ -578,7 +582,10 @@ export default class Xapi extends XapiBase {
       )
     )
 
-    return vm
+    return {
+      size: sizeStream.size,
+      vm
+    }
```
julien-f: Do not change this method, move this in `xo-mixins/backups`.

PR #550, opened 2017-05-19
feat(VM): ability to block/unblock operations

badrAZ:
See vatesfr/xo-web#2119


src/api/vm.coffee
```diff
@@ -909,6 +907,21 @@ exports.start = start
...
+
+forceStart.params = {
+  id: { type: 'string' }
+}
+
+forceStart.resolve = {
+  vm: ['id', 'VM', 'operate']
+}
+
+exports.forceStart = forceStart
```
badrAZ: I will add force to vm.start, it's better.

src/xapi/index.js
```diff
@@ -1437,9 +1437,17 @@ export default class Xapi extends XapiBase {
...
 
-  async startVm (vmId) {
+  async startVm (vmId, force) {
+    const vm = this.getObject(vmId)
+
+    if (force) {
+      await this._updateObjectMapProperty(vm, 'blocked_operations', {
+        start: null
+      })
+    }
```
julien-f: Use reactive workflow instead of conditional workflow.

src/xapi/index.js
```diff
@@ -1437,9 +1437,17 @@ export default class Xapi extends XapiBase {
     )
   }
 
-  async startVm (vmId) {
+  async startVm (vmId, force) {
```
julien-f: Update `_startVm` instead.

PR #556, opened 2017-05-29
feat(xapi-object-to-xo/host): patchesRequiringReboot

pdonias:
See vatesfr/xo-web#2124

### DIFF
--- a/src/xapi-object-to-xo.js
+++ b/src/xapi-object-to-xo.js
@@ -195,6 +195,7 @@ const TRANSFORMS = {
         }),
       agentStartTime: toTimestamp(otherConfig.agent_start_time),
       tags: obj.tags,
+      patchesRequiringReboot: link(obj, 'updates_requiring_reboot'),
       version: obj.software_version.product_version,
 
       // TODO: dedupe.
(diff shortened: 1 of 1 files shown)


src/xapi-object-to-xo.js
```diff
@@ -195,6 +195,7 @@ const TRANSFORMS = {
         }),
       agentStartTime: toTimestamp(otherConfig.agent_start_time),
       tags: obj.tags,
+      patchesRequiringReboot: link(obj, 'updates_requiring_reboot'),
```
julien-f: Shouldn't we call it `updatesRequiringReboot`?

Because IMHO it concerns more than just patches.

pdonias: True. But should xo-web be aware of the difference between patch and update?

olivierlambert: I think it's far too early to decide on this right now. We should merge this and change in the future if necessary.

julien-f: `rebootRequired: !isEmpty(obj.updates_requiring_reboot)`.

PR #558, opened 2017-05-30
feat(backups): block the start operation after the DR backup.

badrAZ:
Fixes vatesfr/xo-web#2097


src/xo-mixins/backups.js
```diff
@@ -1031,6 +1031,10 @@ export default class {
       nameLabel: copyName
     })
 
+    targetXapi._updateObjectMapProperty(data.vm, 'blocked_operations', {
+      start: 'If you want to start this vm, you need to click on Force start.'
```
julien-f: The message should not know anything about the interface.

It can appear in `xo-cli`, `xe` or even `XenCenter`, it does not make sense to have messages too much related to `xo-web`.

julien-f:
To merge after release.

PR #560, opened 2017-05-31
feat(host): update patches when joining pool

pdonias:
See vatesfr/xo-web#878


src/xapi/mixins/patching.js
```diff
@@ -389,6 +390,29 @@ export default {
       : this.installPoolPatchOnHost(patch.uuid, host)
   },
 
+  async installSpecificPatchesOnHost (patchNames, hostId) {
```
julien-f: Use UUIDs or add a `TODO` comment.

PR #567, opened 2017-06-06
feat: ability to define a sr for each VDIs

badrAZ:
See vatesfr/xo-web#2070    vatesfr/xo-web#2180


src/api/vm.coffee
```diff
@@ -726,13 +726,19 @@ exports.rollingDeltaBackup = rollingDeltaBackup
...
+    forEach mapVdisSrs, (srId, vdiId) =>
+      mapVdisSrsXapi[vdiId] = @getObject(srId, 'SR')._xapiId
+
+  return @importDeltaVmBackup({sr, remoteId: remote, filePath, mapVdisSrs: mapVdisSrsXapi})
 
 importDeltaBackup.params = {
   sr: { type: 'string' }
   remote: { type: 'string' }
   filePath: { type: 'string' }
+  mapVdisSr: { type: 'object', optional: true }
```
pdonias: `mapVdisSrs`

src/xapi/index.js
```diff
@@ -995,7 +996,7 @@ export default class Xapi extends XapiBase {
             [TAG_BASE_DELTA]: undefined,
             [TAG_COPY_SRC]: vdi.uuid
           },
-          sr: sr.$id
+          sr: (mapVdisSrs && mapVdisSrs[vdi.uuid]) || sr.$id
```
pdonias: With your implementation, won't `mapVdisSrs` always be defined?

badrAZ: yes, mapVdisSrs is optional. The user can choose  the default sr for all his vdis.

pdonias: Ok but in `vm.coffee` you create a `mapVdisSrsXapi` object even if `mapVdisSrs` is undefined. So if the user chose the default SR for all the VDIs, `mapVdisSrsXapi` will be an empty object, but not undefined.

badrAZ: yes but mapVdisSrsXapi[vdi.uuid] will be undefined.

src/xapi/index.js
```diff
@@ -995,7 +996,7 @@ export default class Xapi extends XapiBase {
             [TAG_BASE_DELTA]: undefined,
             [TAG_COPY_SRC]: vdi.uuid
           },
-          sr: sr.$id
+          sr: mapVdisSrs[vdi.uuid] || sr.$id
```
julien-f: What if `mapVdiSrs` is `undefined`?

badrAZ: In API#VM mapVdisSrs receive an empty object if it is undefined.

src/api/vm.coffee
```diff
@@ -726,13 +726,19 @@ exports.rollingDeltaBackup = rollingDeltaBackup
 
 #---------------------------------------------------------------------
 
-importDeltaBackup = ({sr, remote, filePath}) ->
-  return @importDeltaVmBackup({sr, remoteId: remote, filePath})
+importDeltaBackup = ({sr, remote, filePath, mapVdisSrs}) ->
+  mapVdisSrsXapi = {}
+
+  forEach mapVdisSrs, (srId, vdiId) =>
+    mapVdisSrsXapi[vdiId] = @getObject(srId, 'SR')._xapiId
```
julien-f: I think we can skip this translation if we make it clear that `mapVdisSrs` contains only UUIDs.

@pdonias what's your opinion?

src/xapi/index.js
```diff
@@ -924,9 +924,10 @@ export default class Xapi extends XapiBase {
   @deferrable.onFailure
   async importDeltaVm ($onFailure, delta, {
     deleteBase = false,
+    disableStartAfterImport = true,
+    mapVdisSrs = {},
     name_label = delta.vm.name_label,
-    srId = this.pool.default_SR,
-    disableStartAfterImport = true
+    srId = this.pool.default_SR
```
julien-f: I think this param should be optional and works even if no default SR is defined if `mapVdisSrs` is completely defined.

src/xo-mixins/backups.js
```diff
@@ -879,7 +879,7 @@ export default class {
     }
   }
 
-  async importDeltaVmBackup ({sr, remoteId, filePath}) {
+  async importDeltaVmBackup ({sr, remoteId, filePath, mapVdisSrs}) {
```
julien-f: `sr` should now be optional.

src/xapi/index.js
```diff
@@ -947,7 +948,7 @@ export default class Xapi extends XapiBase {
       }
     }
 
-    const sr = this.getObject(srId)
+    const sr = mapVdisSrs === {} && this.getObject(srId || this.pool.default_SR)
```
julien-f: `mapVdisSrs === {}` cannot work, can you tell me why?

src/xo-mixins/backups.js
```diff
@@ -879,12 +879,12 @@ export default class {
...
   }
 
-  async importDeltaVmBackup ({sr, remoteId, filePath}) {
+  async importDeltaVmBackup ({sr, remoteId, filePath, mapVdisSrs = {}}) {
     filePath = `${filePath}${DELTA_BACKUP_EXT}`
     const { datetime } = parseVmBackupPath(filePath)
 
     const handler = await this._xo.getRemoteHandler(remoteId)
-    const xapi = this._xo.getXapi(sr)
+    const xapi = this._xo.getXapi(sr || mapVdisSrs[Object.keys(mapVdisSrs)[0]])
```
julien-f: This is really inefficient, here's what you could do:
```js
const getFirstKey = object => {
  for (const key in object) {
    if (Object.prototype.has.call(object, key)) {
      return key
    }
  }
}
```

src/xo-mixins/backups.js
```diff
@@ -918,7 +918,8 @@ export default class {
 
       vm = await xapi.importDeltaVm(delta, {
         disableStartAfterImport: false,
-        srId: sr._xapiId
+        srId: sr._xapiId,
```
julien-f: `sr` is no longer mandatory.

src/xo-mixins/backups.js
```diff
@@ -129,6 +129,14 @@ async function checkFileIntegrity (handler, name) {
   //  await eventToPromise(stream, 'finish')
 }
 
+const getFirstKey = object => {
+  for (const key in object) {
+    if (Object.prototype.hasOwnProperty.call(object, key)) {
+      return key
+    }
+  }
+}
```
julien-f: Move this in `utils` or its own module to make it reusable.

src/xapi/index.js
```diff
@@ -1000,7 +999,7 @@ export default class Xapi extends XapiBase {
             [TAG_BASE_DELTA]: undefined,
             [TAG_COPY_SRC]: vdi.uuid
           },
-          sr: sr.$id
+          sr: mapVdisSrs[vdi.uuid] || this.getObject(srId || this.pool.default_SR).$id
```
julien-f: I think it's safe to use `srId` and `this.pool.default_SR` directly.

src/xapi/index.js
```diff
@@ -1094,6 +1093,7 @@ export default class Xapi extends XapiBase {
 
   async _migrateVmWithStorageMotion (vm, hostXapi, host, {
     migrationNetwork = find(host.$PIFs, pif => pif.management).$network, // TODO: handle not found
+    mainSr,
```
julien-f: `sr`.

src/xapi/index.js
```diff
@@ -1375,6 +1378,7 @@ export default class Xapi extends XapiBase {
     const accrossPools = vm.$pool !== host.$pool
     const useStorageMotion = (
       accrossPools ||
+      mainSr ||
```
julien-f: All these tests should be more explicit: `!== undefined`.

src/xo-mixins/backups.js
```diff
@@ -918,7 +919,8 @@ export default class {
 
       vm = await xapi.importDeltaVm(delta, {
         disableStartAfterImport: false,
-        srId: sr._xapiId
+        srId: sr && sr._xapiId,
```
julien-f: Explicit test.

src/xapi/index.js
```diff
@@ -999,7 +999,7 @@ export default class Xapi extends XapiBase {
             [TAG_BASE_DELTA]: undefined,
             [TAG_COPY_SRC]: vdi.uuid
           },
-          sr: mapVdisSrs[vdi.uuid] || this.getObject(srId || this.pool.default_SR).$id
+          sr: mapVdisSrs[vdi.uuid] || srId || this.pool.default_SR
```
julien-f: Why not using `this.pool.default_SR` as a default value for `srId` in the params once and for all?

src/xapi/index.js
```diff
@@ -1367,27 +1367,27 @@ export default class Xapi extends XapiBase {
   }
 
   async migrateVm (vmId, hostXapi, hostId, {
-    mainSr,
+    sr,
     migrationNetworkId,
-    mapVifsNetworks,
-    mapVdisSrs
+    mapVifsNetworks = {},
+    mapVdisSrs = {}
```
julien-f: `isEmpty` works fine with `undefined`.

src/xapi/index.js
badrAZ: ok i will use it as a default value for srId.

PR #574, opened 2017-06-22
fix(Xapi#CreateVm): wait for the cloned vifs to be correctly deleted

badrAZ:


src/xapi/mixins/vm.js
```diff
@@ -175,6 +175,8 @@ export default {
     // Destroys the VIFs cloned from the template.
     await Promise.all(mapToArray(vm.$VIFs, vif => this._deleteVif(vif)))
 
+    await this._waitObjectState(vm.$id, vm => vm.$VIFs.length === 0)
```
julien-f: I think it's not the best fix: updating the final `waitObjectState` of the function seems more appropriate.

src/xapi/mixins/vm.js
```diff
@@ -227,7 +230,8 @@ export default {
     // wait for the record with all the VBDs and VIFs
     return this._waitObjectState(vm.$id, vm =>
       vm.VBDs.length === nVbds &&
-      vm.VIFs.length === nVifs
+      vm.VIFs.length === nVifs &&
+      intersection(oldVifs, vm.VIFs).length === 0
```
julien-f: Replace with:
```js
await this.barrier('VM', vm.$ref)
return this.getObjectByRef(vm.$ref)
```

PR #576, opened 2017-06-28
prevent createNetwork() from partial failure

nraynaud:


julien-f:
Can you do the same fix for `createBondedNetwork` please? :)

PR #577, opened 2017-06-28
fix: synchronize pools with servers

badrAZ:
Fixes vatesfr/xo-web#2238


src/api/host.coffee
```diff
@@ -121,7 +121,18 @@ exports.stop = stop
 #---------------------------------------------------------------------
 
 detach = ({host}) ->
-  return @getXapi(host).ejectHostFromPool(host._xapiId)
+  xapi = @getXapi(host)
+  yield xapi.ejectHostFromPool(host._xapiId)
```
julien-f: You cannot use yield outside a coroutine (`coroutine`/`yield` = `async`/`await`).

src/api/host.coffee
```diff
@@ -121,7 +121,18 @@ exports.stop = stop
...
+
+  { user: username, password } = xapi._auth
+
+  server = yield this.registerXenServer({
+    host: host.address,
+    password,
+    username
+  })
+
+  yield this.connectXenServer(server.id)
```
julien-f: These last calls probably should be async and in case of errors should not impact `host.detach`.

src/xo-mixins/xen-servers.js
```diff
@@ -309,8 +310,40 @@ export default class {
 
     xapi.xo.install()
 
+    let redirectedToAnExistingServer = false
+    xapi.on('redirect', async url => {
+      const servers = await this.getAllXenServers()
+      const serverExists = some(
+        servers,
+        server => server.host === url.hostname
```
julien-f: `host` is not `hostname`: `host` can include a protocol prefix (`http://` or `https://`) and a port (`:1024`).

Maybe you should expose the URL parsing function from `xen-api` to ease the comparison.

src/xo-mixins/xen-servers.js
```diff
@@ -309,8 +311,40 @@ export default class {
...
+      await this.updateXenServer(id, {host: url.hostname})
+    })
+
     await xapi.connect().then(
-      () => this.updateXenServer(id, { error: null }),
+      async () => {
+        let error = null
+        if (redirectedToAnExistingServer) {
+          error = {
+            code: 'SESSION_AUTHENTICATION_FAILED',
```
julien-f: Do not use a XAPI error with another meaning that the original one.

src/xo-mixins/xen-servers.js
```diff
@@ -309,8 +311,40 @@ export default class {
...
+      }
+
+      await this.updateXenServer(id, {host: url.hostname})
+    })
+
     await xapi.connect().then(
-      () => this.updateXenServer(id, { error: null }),
+      async () => {
+        let error = null
+        if (redirectedToAnExistingServer) {
```
julien-f: You have no guarantee that this function will be called after the `redirect` callback has been fully executed.

src/api/host.coffee
```diff
@@ -120,8 +120,19 @@ exports.stop = stop
...
 
-detach = ({host}) ->
-  return @getXapi(host).ejectHostFromPool(host._xapiId)
+detach = $coroutine ({host}) ->
+  xapi = @getXapi(host)
+  yield xapi.ejectHostFromPool(host._xapiId)
+
+  { user: username, password } = xapi._auth
+
+  server = yield this.registerXenServer({
```
badrAZ: No it's not pertinent to wait for it, i will `catch(noop)` the error.

badrAZ: No, it's not pertinent to wait for it in host.detach.

src/xo-mixins/xen-servers.js
```diff
@@ -322,10 +326,48 @@ export default class {
...
+        await this.updateXenServer(id, {host: url.hostname, force: true})
+        resolveRedirectedToAnExistingServer(false)
+      }
+    })
+
     await xapi.connect().then(
-      () => this.updateXenServer(id, { error: null }),
+      async () => {
+        let error = null
+        if (redirect && await redirectedToAnExistingServer) {
```
julien-f: What if there is no redirect?

PR #579, opened 2017-06-30
fix(vm.create): do not take template disks into account for limits

pdonias:


pdonias:
See pull request vatesfr/xo-web#2249.

PR #580, opened 2017-07-03
fix(server.set): it should not be possible to set a connected server

badrAZ:


src/api/server.js
```diff
@@ -69,6 +70,13 @@ getAll.permission = 'admin'
 // -------------------------------------------------------------------
 
 export async function set ({id, ...props}) {
+  if (find(
+    await this.getAllXenServers(),
+    { id }
+  ).status === 'connected') {
+    throw new Error('you can only set the server when it is disconnected.')
+  }
```
julien-f: Update `updateXenServer()` instead.

badrAZ: `updateXenServer()` is used  also in `connectXenServer()` to update the error property.

julien-f: If you just do the check when updating certain entries it should be fine.

src/xo-mixins/xen-servers.js
```diff
@@ -98,6 +99,18 @@ export default class {
...
+      host !== undefined ||
+      password !== undefined ||
+      readOnly !== undefined ||
+      username !== undefined
+
+    if (
+      needDisconnectTheServer &&
+      find(await this.getAllXenServers(), { id }).status === 'connected'
+    ) {
+      throw new Error('this entry require disconnecting the server to update it.')
```
julien-f: No leading uppercase and no trailing period in error message.

src/xo-mixins/xen-servers.js
```diff
@@ -98,6 +99,18 @@ export default class {
     username
   }) {
     const server = await this._getXenServer(id)
+    const needDisconnectTheServer = allowUnauthorized !== undefined ||
```
julien-f: `needDisconnectTheServer` → `requireDisconnected`

julien-f: Move `allowUnauthorized !== undefined` on the next line so it's at the same level of other entries.

src/xo-mixins/xen-servers.js
```diff
@@ -98,6 +99,18 @@ export default class {
...
+      host !== undefined ||
+      password !== undefined ||
+      readOnly !== undefined ||
+      username !== undefined
+
+    if (
+      needDisconnectTheServer &&
+      find(await this.getAllXenServers(), { id }).status === 'connected'
```
julien-f: No need to use either `find` or `getAllXenServers`.

src/xo-mixins/xen-servers.js
```diff
@@ -98,6 +98,21 @@ export default class {
     username
   }) {
     const server = await this._getXenServer(id)
+    const xapi = this._xapis[id]
+    const requireDisconnected =
+      allowUnauthorized !== undefined ||
+      host !== undefined ||
+      password !== undefined ||
+      readOnly !== undefined ||
```
julien-f: Can be changed while connected.

PR #584, opened 2017-07-05
fix(Xapi#findAvailableSharedSr): should not be async

pdonias:
Fixes vatesfr/xo-web#2262


src/xapi/index.js
```diff
@@ -1603,6 +1603,10 @@ export default class Xapi extends XapiBase {
     type = 'user',
     xenstore_data = undefined
   } = {}) {
+    if (sr == null) {
```
julien-f: Test against `OPAQUE_REF_NULL`.

PR #586, opened 2017-07-13
fix(vm.create): assign device to VDIs

pdonias:


src/api/vm.coffee
```diff
@@ -82,35 +82,13 @@ create = $coroutine (params) ->
...
-      SR: sr._xapiId,
-      type: vdi.type
-    })
-  )
 
+  highestDevice = 0
+  vdiSizesByDevice = {}
   existingVdis = extract(params, 'existingDisks')
   params.existingVdis = existingVdis and map(existingVdis, (vdi, userdevice) =>
+    highestDevice = Math.max(highestDevice, +userdevice)
```
julien-f: It's not enough, `existingVdis` contains optional changes to apply to the existing disks.

You should check the template's disk to really know the max `device`.

PR #594, opened 2017-08-25
feat(jobs): fail the smart backup job when it's empty

badrAZ:
See vatesfr/xo-web#2253


src/job-executor.js
```diff
@@ -68,7 +77,32 @@ const paramsVectorActionsMap = {
...
+      if (pool !== undefined) {
+        error += `- ${pool.not ? 'Not in' : 'In'} pools: ${pool.values.join(', ')}\n`
+      }
+
+      if (tags !== undefined) {
+        error += `- ${tags.not ? 'Does not contain' : 'Contains'} one of these tags: ${tags.values.join(', ')}`
+      }
+
+      throw new Error(error)
+    }
```
julien-f: Too complicated, throw a generic error.

src/job-executor.js
```diff
@@ -68,7 +71,11 @@ const paramsVectorActionsMap = {
     ))
   },
   fetchObjects ({ pattern }) {
-    return filter(this.xo.getObjects(), object => match(pattern, object))
+    const objects = filter(this.xo.getObjects(), object => match(pattern, object))
+    if (isEmpty(objects)) {
+      throw new Error('No VMs respect the smart backup conditions!')
```
julien-f: This code is not directly related to smart backups and is not specific to VMs.

julien-f:
@pdonias what do you think about this PR now?

PR #605, opened 2017-09-25
feat(self): handle add/remove VDI on VM

pdonias:
See vatesfr/xo-web#2348

### DIFF
--- a/src/api/disk.js
+++ b/src/api/disk.js
@@ -1,12 +1,39 @@
-import {parseSize} from '../utils'
+import { parseSize } from '../utils'
+import { unauthorized } from 'xo-common/api-errors'
 
 // ===================================================================
 
-export async function create ({name, size, sr}) {
-  const vdi = await this.getXapi(sr).createVdi(parseSize(size), {
+export async function create ({ name, size, sr, vm, params }) {
+  const { resourceSet } = vm
+
+  const attach = vm !== undefined
+
+  if (attach && resourceSet != null) {
+    await this.checkResourceSetConstraints(resourceSet, this.user.id, [ sr ])
+    await this.allocateLimitsInResourceSet({ disk: size }, resourceSet)
+  } else {
+    if (!(await this.hasPermissions(this.session.get('user_id'), [ [ sr, 'administrate' ] ]))) {
+      throw unauthorized()
+    }
+  }
+
+  const xapi = this.getXapi(sr)
+  sr = this.getObject(sr, 'SR')
+
+  const vdi = await xapi.createVdi(parseSize(size), {
     name_label: name,
     sr: sr._xapiId
   })
+
+  if (attach) {
+    const { bootable, position, mode } = params
+    await xapi.attachVdiToVm(vdi.$id, vm._xapiId, {
+      bootable,
+      position,
+      readOnly: mode === 'RO'
+    })
+  }
+
   return vdi.$id
 }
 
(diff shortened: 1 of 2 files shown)


src/api/disk.js
```diff
@@ -1,12 +1,39 @@
-import {parseSize} from '../utils'
+import { parseSize } from '../utils'
+import { unauthorized } from 'xo-common/api-errors'
 
 // ===================================================================
 
-export async function create ({name, size, sr}) {
-  const vdi = await this.getXapi(sr).createVdi(parseSize(size), {
+export async function create ({ name, size, sr, vm, params }) {
+  const { resourceSet } = vm
```
julien-f: What if the VM is `undefined`?

src/api/disk.js
```diff
@@ -1,12 +1,38 @@
...
-  const vdi = await this.getXapi(sr).createVdi(parseSize(size), {
+export async function create ({ name, size, sr, vm, params }) {
+  const attach = vm !== undefined
+
+  let resourceSet
+  if (attach && (resourceSet = vm.resourceSet) != null) {
+    await this.checkResourceSetConstraints(resourceSet, this.user.id, [ sr ])
+    await this.allocateLimitsInResourceSet({ disk: size }, resourceSet)
+  } else {
+    if (!(await this.hasPermissions(this.session.get('user_id'), [ [ sr, 'administrate' ] ]))) {
```
julien-f: `else`+`if` → `else if`

src/api/disk.js
```diff
@@ -1,12 +1,38 @@
...
+      throw unauthorized()
+    }
+  }
+
+  const xapi = this.getXapi(sr)
+  sr = this.getObject(sr, 'SR')
```
julien-f: What's the point?
Use `resolve` instead.

pdonias: How do you specify that the user doesn't need ACLs on the SR with `resolve`? I thought I could do it with `sr: [ 'sr', 'SR', '' ]` but still got a "not authenticated or not enough permissions" error when I tried it.

julien-f: That's weird, you shouldn't.

Passing `undefined` default to `administrate`, but using `false` (or any falsy value), should simply bypass the ACLs.

src/api/disk.js
```diff
@@ -1,12 +1,34 @@
...
-export async function create ({name, size, sr}) {
-  const vdi = await this.getXapi(sr).createVdi(parseSize(size), {
+export async function create ({ name, size, sr, vm, params }) {
+  const attach = vm !== undefined
+
+  let resourceSet
+  if (attach && (resourceSet = vm.resourceSet) != null) {
+    await this.checkResourceSetConstraints(resourceSet, this.user.id, [ sr.id ])
+    await this.allocateLimitsInResourceSet({ disk: size }, resourceSet)
+  } else if (!(await this.hasPermissions(this.session.get('user_id'), [ [ sr.id, 'administrate' ] ]))) {
```
julien-f: `this.user.id`?

src/api/disk.js
```diff
@@ -15,11 +37,22 @@ create.description = 'create a new disk on a SR'
...
+  vm: { type: 'string' },
+  params: {
+    type: 'object',
+    optional: true,
+    properties: {
+      bootable: { type: 'boolean', optional: true },
+      mode: { type: 'string', optional: true },
+      position: { type: 'string', optional: true }
+    }
+  }
```
julien-f: Why not inlining those params?

src/api/disk.js
```diff
@@ -15,11 +37,22 @@ create.description = 'create a new disk on a SR'
...
 }
 
 create.resolve = {
-  sr: ['sr', 'SR', 'administrate']
+  vm: ['vm', 'VM', 'administrate'],
+  sr: ['sr', 'SR', '']
```
julien-f: Don't you thing `false` instead of `''` is clearer?

src/api/vdi.coffee
```diff
@@ -6,10 +6,20 @@
 {invalidParameters, unauthorized} = require 'xo-common/api-errors'
 {isArray: $isArray, parseSize} = require '../utils'
 {JsonRpcError} = require 'json-rpc-peer'
+{reduce} = require 'lodash'
 
 #=====================================================================
 
 delete_ = $coroutine ({vdi}) ->
+  resourceSet = reduce(
```
julien-f: I'm not sure a `reduce` is the most appropriate here.

PR #613, opened 2017-10-20
feat(objects): add gpu group and vgpu type

olivierlambert:
This is a first step to expose GPU features of XenServer. More will come in another PR (API to create a new VGPU into a VM etc.)

### DIFF
--- a/src/xapi-object-to-xo.js
+++ b/src/xapi-object-to-xo.js
@@ -693,8 +693,7 @@ const TRANSFORMS = {
       device: obj.device,
       gpuGroup: link(obj, 'GPU_group'),
       resident_on: link(obj, 'resident_on'),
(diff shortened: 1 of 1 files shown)


src/xapi-object-to-xo.js
```diff
@@ -691,9 +696,49 @@ const TRANSFORMS = {
...
+  gpu_group (obj) {
+    return {
+      type: 'gpuGroup',
+
+      allocation: obj.allocation_algorithm,
+      enabledVgpuTypes: link(obj, 'enabled_VGPU_types'),
+      gpuTypes: obj.GPU_types,
+      name_description: obj.name_description,
+      name_label: obj.name_label,
+      other_config: obj.other_config,
```
julien-f: Use camelCase everywhere.

PR #618, opened 2017-11-13
feat(backups): get the merged data size and its duration for Delta backups

badrAZ:
See vatesfr/xo-web#2426


src/xo-mixins/backups.js
```diff
@@ -893,7 +903,11 @@ export default class {
     return {
       // Returns relative path.
       path: `${dir}/${backupFormat}`,
-      size: dataSize
+      size: dataSize,
+      merge: {
+        size: mergedDataSize,
+        duration: end - start
+      }
```
julien-f: `size` → `transferSize`, `transferDuration`, `mergeSize`, `mergeDuration`.

src/vhd-merge.js
```diff
@@ -540,6 +540,9 @@ class Vhd {
 
       i = endSector
     }
+
+    // Return the merged data size
+    return data.length
```
julien-f: Unnecessary, you can use `this.fullBlockSize` directly.

PR #626, opened 2017-12-11
fix(backupVm): remove XVA on failure

julien-f:
Fixes vatesfr/xo-web#2524


julien-f:
Should be fixed now :)

PR #628, opened 2017-12-12
feat(xosan): license management

pdonias:


src/xo-mixins/xosan.js
```diff
@@ -0,0 +1,62 @@
...
+      throw new Error('no license')
+    })
+  }
+
+  async unbindXosanLicense (srId) {
+    const license = await this._xo.checkLicense({ boundObjectId: srId })
+
+    if (license === undefined) {
+      return
+    }
```
julien-f: This test is useless.

src/xo-mixins/xosan.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  getXosanTrialLicenseId () {
+    return this._xo.getAvailableTrial({ productId: 'xosan' }).then(
+      license => license && license.id
+    )
+  }
+
+  async bindXosanTrialLicense (licenseId, srId) {
+    const license = await this._xo.getLicense({ licenseId, productId: 'xosan.trial' })
```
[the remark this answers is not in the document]
marsaud: There is no licenseId paramater on `getLicense`, only productId so far.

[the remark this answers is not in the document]
marsaud: IMHO getLicenses should not throw, as "no such licenses" is not an exception, we want to get licenses sharing a criteria, and if there are none, we get any empty collection. It falls to the client developer to treat this as an exception or not in the scenario he is implementing.

src/xo-mixins/xosan.js
```diff
@@ -0,0 +1,62 @@
...
+
+    if (license === undefined) {
```
marsaud: All this is strange, what is is this `getLicense` code before calling `useLicense`, you have a `licenseId` from the start ... what do tou mean to do here ?

src/xo-mixins/xosan.js
```diff
@@ -0,0 +1,62 @@
...
+      throw new Error('trial license not found')
+    }
+
+    return this._xo.useLicense({ licenseId, boundObjectId: srId })
+  }
+
+  async getLicense (licenseId, productId = 'xosan') {
```
marsaud: Why do you need this?

src/xo-mixins/xosan.js
```diff
@@ -0,0 +1,62 @@
...
+      if (license.expires < Date.now()) {
+        throw new Error('expired license')
+      }
+    }).catch(() => {
+      throw new Error('no license')
+    })
+  }
+
+  async unbindXosanLicense (srId) {
+    const license = await this._xo.checkLicense({ boundObjectId: srId })
```
marsaud: Strange ... this function looks like a "try to unbind if it happens it is bound". Any scenario leading to break a bound shall follow a fully tested context, so you could call `freeLicense` being sure of what you are doing, and sure of your parameters.

src/xo-mixins/xosan.js
```diff
@@ -0,0 +1,62 @@
...
+
+    if (license === undefined) {
+      return
+    }
+
+    return this._xo.freeLicense({ licenseId: license.id, boundObjectId: srId })
```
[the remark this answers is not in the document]
marsaud: The server will perform a check with the full parameters, this will avoid any concurrent access between the display and action on UI side.

src/xo-mixins/xosan.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  async unlockXosan (licenseId, srId) {
+    const trialLicense = await this._xo.checkLicense({ boundObjectId: srId }) // May throw an error if no license
+    if (trialLicense == null || trialLicense.productId !== 'xosan.trial') {
```
marsaud: You know trialLicense will throw if there is no license bound, so why test if there is a license ?

src/xo-mixins/xosan.js
```diff
@@ -0,0 +1,62 @@
...
+      throw new Error(`bad XOSAN trial license on SR ${srId}`)
+    }
+
+    const freedLicense = await this._xo.freeLicense({ licenseId: trialLicense.id, boundObjectId: srId })
+    if (freedLicense == null || freedLicense.boundObjectId !== undefined) {
+      throw new Error(`could not free license ${trialLicense.id} from SR ${srId}`)
+    }
```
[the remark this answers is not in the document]
marsaud: They are. `freeLicense` cannot return null or undefined and cannot end returning a license object with an assigned boundObjectId.

src/xo-mixins/xosan.js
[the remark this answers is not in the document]
pdonias: This is a typo. I meant to call `this.getLicense` here, not `this._xo.getLicense` which doesn't even exist. I need to pass the `productId` because I want to check if we're actually trying to bind an XOSAN trial license here. This is just a protection.

src/xo-mixins/xosan.js
pdonias: Without the typo, I want to call `this.getLicense(licenseId, 'xosan.trial')` here to check if we're actually trying to bind a trial license to the freshly created XOSAN SR. This should never be anything else but it's just a protection to prevent binding an incorrect license.

src/xo-mixins/xosan.js
pdonias: I use it twice:
- before binding a new XOSAN to its trial license to check if the trial license actually exists
- before unlocking a trial XOSAN with a paid license to check if the license exists and is not expired before starting the process of *unbinding the trial license* then *binding the paid one*

src/xo-mixins/xosan.js
pdonias: This is precisely a "try to unbind" method. For instance, I'm calling this when deleting a SR (XOSAN or classic). If the SR happens to be bound to a license, it will unbind it.
Do you think this behaviour will lead to some issues?

src/xo-mixins/xosan.js
```diff
@@ -0,0 +1,62 @@
...
+    return this._xo.getAvailableTrial({ productId: 'xosan' }).then(
+      license => license && license.id
+    )
+  }
+
+  async bindXosanTrialLicense (licenseId, srId) {
+    const license = await this._xo.getLicense({ licenseId, productId: 'xosan.trial' })
+
+    if (license === undefined) {
+      throw new Error('trial license not found')
```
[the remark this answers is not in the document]
marsaud: But you have claimed this trial license yourself, that's how you know it's id. How can you think it can be another license than it is ? The productId of a license never changes.

src/xo-mixins/xosan.js
```diff
@@ -0,0 +1,62 @@
...
+
+    return this._xo.useLicense({ licenseId, boundObjectId: srId })
+  }
+
+  async getLicense (licenseId, productId = 'xosan') {
+    return find(await this._xo.getLicenses({ productId }), [ 'id', licenseId ])
+  }
+
+  checkXosanLicense (sr) {
+    const srId = (sr && sr.id) || sr
```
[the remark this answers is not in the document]
marsaud: I don't follow you, you have all information on all of the customer's licenses through `getLicenses`, you indeed use this information to let the client choose his actions and prepare your calls in consequence. If something has changed on server side meanwhile, you will recieve explicit erros, as explicit as what you can do yourself by making thess "over-tests", no ?

src/xo-mixins/xosan.js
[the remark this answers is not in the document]
marsaud: Ah ok, I see. Well, I would have called it "forceUnbindXosanLicense", as to me, a simple unbind shall throw if there is no license to unbind.

src/api/xosan.js
```diff
@@ -110,46 +109,92 @@ export async function getVolumeInfo ({sr, infoType}) {
...
+    profile: profileType,
+    profileTop: profileTopType,
+    hosts: checkHosts,
   }
+}
+
+const VOLUME_INFO_TYPES = createVolumeInfoTypes()
+
+export async function getVolumeInfo ({sr, infoType}) {
+  await this.checkXosanLicense(sr)
```
julien-f: Is SR a UUID or an opaque ref?

Because almost all xo-server's API code is currently accepting both indifferently.

src/api/xosan.js
```diff
@@ -1011,6 +1117,25 @@ computeXosanPossibleOptions.params = {
...
+  await glusterEndpoint.xapi.call('SR.scan', glusterEndpoint.xapi.getObject(srId).$ref)
+}
+
+unlock.description = 'Unlock XOSAN SR functionalities by binding it to a paid license'
+
+unlock.permission = 'admin'
+
+unlock.params = {
+  licenseId: { type: 'string' },
+  srId: { type: 'string' },
```
julien-f: Why is it `srId` while everywhere else it's just  `sr`?

PR #646, opened 2018-01-11
fix(proxy-console): socket.close() → socket.end()

pdonias:


olivierlambert:
what was the impact of this issue?

pdonias:
xo-server crashed on console's websocket error.

olivierlambert:
ouch! is it an old bug?

pdonias:
Sep 2015, probably.

olivierlambert:
Strange we never noticed it before. Thanks for the fix!

PR #648, opened 2018-01-30
feat(vm/delete): add force param to bypass blocked_operations

pdonias:
See vatesfr/xo-web#2525


src/api/vm.js
```diff
@@ -322,7 +322,7 @@ create.resolve = {
 
 // -------------------------------------------------------------------
 
-async function delete_ ({ vm, delete_disks: deleteDisks = false }) {
+async function delete_ ({ vm, delete_disks: deleteDisks = false, force = false }) {
```
julien-f: IMHO the default value should be specified only once in `Xapi`.

PR #649, opened 2018-01-31
feat(vm.set): move a vm in to/out of a Self Service Group

badrAZ:
See vatesfr/xo-web#1913


src/api/vm.js
```diff
@@ -498,11 +498,60 @@ migrate.resolve = {
...
+    const newResourceSet = newResourceSetId && await this.getResourceSet(newResourceSetId)
+    const previousResourceSet = previousResourceSetId && await this.getResourceSet(previousResourceSetId)
+    if (previousResourceSetId === undefined) {
+      await updateResourceSet.call(this, VM, newResourceSetId, this.allocateLimitsInResourceSet.bind(this))
+      xapi.xo.setData(vmId, 'resourceSet', newResourceSetId)
+      updateAcls(vmId, newResourceSet, this.addAcl)
+    } else if (newResourceSetId === null) {
+      await updateResourceSet.call(this, VM, previousResourceSetId, this.releaseLimitsInResourceSet.bind(this))
+      xapi.xo.setData(vmId, 'resourceSet', newResourceSetId)
+      updateAcls(vmId, previousResourceSet, this.removeAcl)
```
julien-f: Mutualize this case with `vm.delete`.

src/api/vm.js
```diff
@@ -498,11 +499,60 @@ migrate.resolve = {
...
+  const newResourceSetId = extract(params, 'resourceSet')
+
+  if (newResourceSetId !== undefined) {
+    if (this.user.permission !== 'admin') {
+      throw unauthorized()
+    }
+
+    const previousResourceSetId = xapi.xo.getData(vmId, 'resourceSet')
+    if (previousResourceSetId === undefined && newResourceSetId === null && size(params) === 0) {
+      console.warn('the vm is not in a resource set')
```
julien-f: Not an error, should behave as expected.

src/api/vm.js
```diff
@@ -498,11 +499,60 @@ migrate.resolve = {
...
+      return
```
julien-f: Avoid return here, only handle resource set related stuff.

src/xo-mixins/resource-sets.js
```diff
@@ -344,4 +345,67 @@ export default class {
 
     await Promise.all(mapToArray(sets, set => this._save(set)))
   }
+
+  async allocateVmLimitsInResourceSet (vm, resourceSetId) {
```
julien-f: Private method: `_allocateVmLimits`.

src/xo-mixins/resource-sets.js
```diff
@@ -344,4 +345,67 @@ export default class {
...
+    const resourceSetUsage = this.computeVmResourcesUsage(vm)
+    const ipPoolsUsage = await this._xo.computeVmIpPoolsUsage(vm)
+    return this.allocateLimitsInResourceSet(
+      merge(resourceSetUsage, ipPoolsUsage),
```
julien-f: Is `merge` necessary compared to `assign`?

src/xo-mixins/resource-sets.js
```diff
@@ -344,4 +345,67 @@ export default class {
...
+  }
+
+  async releaseVmLimitsInResourceSet (vm, resourceSetId) {
+    const resourceSetUsage = this.computeVmResourcesUsage(vm)
+    const ipPoolsUsage = await this._xo.computeVmIpPoolsUsage(vm)
+    return this.releaseLimitsInResourceSet(
+      merge(resourceSetUsage, ipPoolsUsage),
+      resourceSetId
+    )
+  }
```
julien-f: We should find a way to merge these two methods.

Maybe we should simply integrate `computeVmIpPoolsUsage` into `computeVmResourcesUsage`?

src/xo-mixins/resource-sets.js
```diff
@@ -344,4 +345,67 @@ export default class {
...
+  async allocateVmLimitsInResourceSet (vm, resourceSetId) {
+    const resourceSetUsage = this.computeVmResourcesUsage(vm)
+    const ipPoolsUsage = await this._xo.computeVmIpPoolsUsage(vm)
+    return this.allocateLimitsInResourceSet(
+      merge(resourceSetUsage, ipPoolsUsage),
+      resourceSetId
+    )
+  }
+
+  async releaseVmLimitsInResourceSet (vm, resourceSetId) {
```
julien-f: Private method: `_deallocateVmLimits`.

src/xo-mixins/resource-sets.js
```diff
@@ -344,4 +345,67 @@ export default class {
...
+    const resourceSetUsage = this.computeVmResourcesUsage(vm)
+    const ipPoolsUsage = await this._xo.computeVmIpPoolsUsage(vm)
+    return this.releaseLimitsInResourceSet(
+      merge(resourceSetUsage, ipPoolsUsage),
+      resourceSetId
+    )
+  }
+
+  async addVmToResourceSet (vm, resourceSetId) {
```
julien-f: Public methods should receive object ids, not object directly.

Look at the existing methods.

src/xo-mixins/resource-sets.js
```diff
@@ -344,4 +345,67 @@ export default class {
...
+    await this.allocateVmLimitsInResourceSet(vm, resourceSetId)
+
+    const xapi = this._xo.getXapi(vm)
+    const vmId = vm._xapiId
+    xapi.xo.setData(vmId, 'resourceSet', resourceSetId)
+
+    const resourceSet = await this.getResourceSet(resourceSetId)
+    forEach(resourceSet.subjects, subjectId =>
+      this._xo.addAcl(subjectId, vmId, 'admin')
```
julien-f: This is an async method, you should wait for them to terminate properly.

src/xo-mixins/resource-sets.js
badrAZ: it used in `vm.delete`

julien-f: Why not using `removeVmFromResourceSet`?

badrAZ: because the vm will be deleted so we don't need to remove `resourceSet` property and the ACLs

src/xo-mixins/resource-sets.js
badrAZ: i don't think so.

julien-f: If not necessary, replace by `assign` :)

src/xo-mixins/resource-sets.js
julien-f: If you don't to wait, ignore properly the errors: `this._xo.addAcl(subjectId, vmId, 'admin')::ignoreErrors`.

